PR #13, opened 2017-01-04
add some basic npm scripts

Chris-R3:


stubailo:
Good idea, thanks! That way people don't need webpack globally installed.

PR #14, opened 2017-01-04
add basic support for devtools dark theme

Chris-R3:
I changed some text colour's to make them more readable when the dark theme is active

### DIFF
--- a/app/components/Inspector/inspector.less
+++ b/app/components/Inspector/inspector.less
@@ -34,6 +34,11 @@
     border-bottom: 1px solid #CCC;
     line-height: 33px;
     padding-left: 10px;
+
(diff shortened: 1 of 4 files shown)


Jpunt:
Seems to work great! Some nitpicky details:


- There seems to be some weird flickering in the store-view (and only in that view): https://dl.dropboxusercontent.com/u/661839/devtools.mov

- I'd love to have the same (default) colors as in my console. The colors in the Apollo-pane seem to be pretty hard to read with a dark background:

screen shot 2017-01-16 at 11 53 09

daniman:
This looks good, thanks @Chris-R3! I'll add some follow-up CSS regarding @Jpunt and make another release.

PR #19, opened 2017-02-08
fix: add typecheck to logger to allow redux-thunk

mmissey:
For actions that return redux-thunks, Apollo-client-devtools middleware was throwing a breaking error `Cannot call function "split" of undefined`. 
This type check avoids any error where type does not exist or is not a string.
[example](https://cloud.githubusercontent.com/assets/563568/22762129/0b47a8f4-ee1b-11e6-8281-73051e9b9dd9.png)

### DIFF
--- a/app/components/Panel.js
+++ b/app/components/Panel.js
@@ -90,7 +90,8 @@ export default class Panel extends Component {
 
           const logger = (logItem) => {
             // Only log Apollo actions for now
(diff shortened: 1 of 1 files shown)


daniman:
This is great, thanks! I'll merge it and release v1.0.5 to the Chrome Webstore.

Issue #23, opened 2017-03-08
Not working in Electron


jasonbahl:
I've got React, Redux and Devtron devtools working in Electron, but can't get apollo-client-devtools working. 

I'm using `electron-devtools-installer` and the following: 

```
installer.default('jdkknkkbebbapilgoeccciglkfbmbnfm', true )
      .then((name) => console.log(name, ' INSTALLED\n'))
      .catch((err) => console.log('An error occurred: ', err))
```

Outputs: 

```
Apollo Client Developer Tools  INSTALLED
```

I'm using the same tool to download the other devtools as well. 

However, when I open Electron devtools, I only see React, Redux and Devtron devtools, no sign of Apollo devtools. 

I followed the logic and it looks like it's supposed to only instantiate if `window.__APOLLO_CLIENT__` is set, and `process.env.NODE_ENV` is not production, and both are the case for me. . .but I still get no Apollo dev tools. 

I have a pretty similar setup (same React/Apollo setup) in just a simple JS web app, and Apollo Dev Tools are present in Chrome, but just not in Electron.

I'll try and dig and see if I can find anything.

jasonbahl:
Ok, so I did some digging and based on the Electron docs here (https://github.com/electron/electron/blob/master/docs/tutorial/devtools-extension.md) I was able to get Apollo Dev Tools working by adding the following: 

`BrowserWindow.addDevToolsExtension( '~/Library/Application Support/Google/Chrome/Default/Extensions/jdkknkkbebbapilgoeccciglkfbmbnfm/1.0.5_0' );`

The other dev-tools work without having to specify this in my code, but adding this for Apollo Dev Tools got it working. 

I'm not super familiar with the Dev Tools API's, but I'm curious if there's something that could be changed so that the Dev tools are instantiated, like the other dev tools, without the explicit call being added.

I have a feeling it *might* be related to how the panel is instantiated here: https://github.com/apollographql/apollo-client-devtools/blob/master/extension/devtools.js

It seems like the other dev tools create the panel without any conditions being checked, and check conditions inside the panel itself, such as Redux Dev Tools: 
https://github.com/zalmoxisus/redux-devtools-extension/blob/master/src/browser/extension/devtools/index.js

jasonbahl:
Ah. . .

[...]

PR #24, opened 2017-03-08
Dev Tools Loading

jasonbahl:
This adjusts the loading of the Dev Tools panel to load later if Apollo is added after the page load.

This addresses #23

### DIFF
--- a/extension/devtools.js
+++ b/extension/devtools.js
@@ -1,10 +1,32 @@
-chrome.devtools.inspectedWindow.eval(
-  `!!(window.__APOLLO_CLIENT__)`,
-   function(result, isException) {
-     if (result) {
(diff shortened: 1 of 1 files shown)


intellix:
Also having issues with the DevTools tab not showing at all. I'm using Angular with Angular CLI. Apollo is loaded lazily via Dependency Injection so the `window.__APOLLO_CLIENT__` property isn't even set until I visit a page that injects or uses it.

If I load the page that uses Apollo, the tab still doesn't appear. Hope this PR fixes it

jasonbahl:
I'll leave this up to the maintainers whether it's worth merging still or not, but I'm not convinced it solves the issue I opened the PR trying to solve. See my latest note on #23 and what I'm doing now.

stubailo:
I'm going to close this because it's not clear if it solves the relevant issue.

Issue #18, opened 2017-01-18
Show GraphQL Errors


SachaG:
My main use case for Redux Devtools when working with Apollo is for viewing GraphQL query errors (which are not surfaced as regular JavaScript errors and also don't get thrown with any meaningful information on the server):

![https://d3vv6lp55qjaqc.cloudfront.net/items/2R2P1y0P0S1K1J3o2c3p/Screen%20Shot%202017-01-18%20at%2010.28.04.png?X-CloudApp-Visitor-Id=9292ee42fa5c7b5a4aea538793ce9914&v=46272b7f](https://d3vv6lp55qjaqc.cloudfront.net/items/2R2P1y0P0S1K1J3o2c3p/Screen%20Shot%202017-01-18%20at%2010.28.04.png?X-CloudApp-Visitor-Id=9292ee42fa5c7b5a4aea538793ce9914&v=46272b7f)

Since displaying these errors using the Redux Devtools requires drilling down multiple levels, it'd be super helpful if the Apollo Devtools could surface them in an easier way. As far as I can tell there's currently no way to see errors, even in the queries pane.

stubailo:
I think it should be quite easy to display the errors in the queries pane since it's just a matter of reading from the store. Might be an easy PR, want to take a stab at it?

SachaG:
I'm a bit busy right now so I don't want to claim this if somebody else wants to work on it, but I'll let you know if I do find the time. I do have a few ideas on what I'd like to see (for example, highlight the queries tab in red if a query has errors).

stubailo:
Yeah that would be a great improvement.

daniman:
This has been added, thanks to #25!

SachaG:
Oh that's great! Nice work :)

PR #25, opened 2017-03-21
Display GraphQL / network errors in the Queries panel

bryan-coursera:
This commit adds two new sections to the `WatchedQuery` component: GraphQL errors and network errors. This data is pulled directly off the query prop object.
This also displays a small warning icon in the query list, to easily detect which queries failed, and fixes a small bug with the `LabeledShowHide` component that wasn't respecting the initial show prop.

I think this addresses #18.

Preview:
devtools

Note: I took the warning icon from https://github.com/percolatestudio/percolate-icons/blob/master/v3.2/SVG/alert.svg, which is where I think some of the icons in Apollo Optics are from (and I think MDG recently acquired Percolate Studio?). Let me know if that's ok, or if I should find an icon from somewhere else.


daniman:
This is great, thanks so much for the PR! Going to merge it and release a new version of the devtools in the next day :)

PR #27, opened 2017-03-27
Fix error running in electron app.

okland:
Fix error while used within Electron app - in which chrome.runtime.getManifest() is not defined.

screen shot 2017-03-27 at 10 12 36


daniman:
Thanks for this! Will release a new version to the Chrome Webstore with the patch today.

PR #41, opened 2017-05-31
CSS fix and added a devtool Icon

maz-dev:
Added a white background on some panels in order to support Firefox Dark Theme and I added a white icon, firefox is automatically handling the color inversion on selected/unselected


maz-dev:
Firefox extension is already usable as is in V54. This version will be released on the main channel (the one used by firefox stable) on 2017-06-13. So it might be a good idea to wait until V54 is officially released before publishing Apollo devtools on the firefox addon store. I'll check if everything is still good to go on the day of release and get back to you.
This PR is mostly 4 lines of CSS, an inverted color icon and the corrected icon path in the devtool panel creation function.

You can try this in about:debugging in firefox, and by choosing the manifest.json file in the load module file picker.

I didn't commit the bundle.js file.

Cheers.

maz-dev:
I can confirm this works with the stable firefox version.

PR #43, opened 2017-06-11
[WIP] Push-based data flow

rrdelaney:
**Still a work in-progress, not ready to merge**

This sets the basis for a push-based data flow to the devtools. Instead of running eval-ing the store and adding it to an action queue, this inserts middleware to dispatch all Apollo client related Redux actions to an Apollo client located in  the devtools. This gives a locally accessible and synced view of the Apollo state in the devtools, so no need to run `evalInPage`.

Still working on converting components to use the local client.

Looking for feedback on the general approach!


stubailo:
We're hoping to decouple the client from Redux soon, perhaps we could come up with an approach that works in a way other than actions? Or maybe it's a fine start for now, and we can solve that problem when we get there.

rrdelaney:
@stubailo 😮 why are you guys planning to move away from Redux?

I can look into another approach in that case

stubailo:
> why are you guys planning to move away from Redux?

We're not moving away entirely, we just want to be able to integrate with a variety of data systems, so we won't be able to rely on Redux long term in tools like this.

stubailo:
Although, this might be a reasonable start since it's going to be a while until we change that!

rrdelaney:
Is there a better way to keep two Apollo caches in sync?

stubailo:
Not yet. I am a bit worried about that aspect of the approach because it's very likely that the devtools will be running a different version than the app itself, right? I don't know if syncing only the actions is going to work in the case of small changes.

stubailo:
(By the way I'm not trying to block anything, we should absolutely merge in something that works and makes an improvement in the short term, even if there might be changes later)

I think @ramyanaga is going to take a look at this and also make some improvements, we can work together!

rrdelaney:
Let me know what I can do to help move this along 😄 A lot of problems here seem to stem from a pull-based approach, and pulling a lot of data. Doing incremental pushes would help a lot I think

rrdelaney:
Here's the basic model I followed when implementing push from content scripts to the devtools panel for reference: https://developer.chrome.com/extensions/devtools#content-script-to-devtools

stubailo:
Closing in favor of #65, thanks for getting us started @rrdelaney!

rrdelaney:
Awesome! Glad I was able to help 😄

Issue #28, opened 2017-04-01
Devtools Completely Locking My Browser


cbranch101:
On initial load, everything seems to work fine, but after a clicking around a couple of times, I start getting an unending stream of long process warnings for processes taking around 250 ms, about twice a second, which completely locks up the browser tab I have open.  Disabling Devtools causes it to go away. 

If there's anything strange about our setup that might be triggering this, I'm guessing it's the fact that we're using deeply nested fragments. 

Here's an example query. 

```
query Section {
  section {
    ...SectionHeaderSection
    ...SectionOutcomesSection
    ...SingleStudentSection
    ...StudentsMonitoringSection
    ...AppBody
    ...RouteForUser
    ...LastUpdatedSection
    __typename
  }
}
fragment SectionHeaderSection on SectionData {
  ...CourseSettingsSection
  ...AverageGradeSummarySection
  ...StudentDistributionSummarySection
  ...SectionInfoSection
  __typename
}
fragment CourseSettingsSection on SectionData {
  stats {
    averages {
      grades
      __typename
    }
    gradesStddev
    __typename
  }
  settings {
    values {
      ...SettingsMessages
      ranges
      __typename
    }
    defaults {
      ...SettingsMessages
      __typename
    }
    gradeHistogram
    __typename
  }
  __typename
}
fragment SettingsMessages on CourseSettingsFields {
  messageLow
  messageMiddle
  messageHigh
  __typename
}
fragment AverageGradeSummarySection on SectionData {
  stats {
    averages {
      grades
      __typename
    }
    __typename
  }
  __typename
}
fragment StudentDistributionSummarySection on SectionData {
  stats {
    ranges {
      grades {
        low {
          ...RangeValues
          __typename
        }
        middle {
          ...RangeValues
          __typename
        }
        high {
          ...RangeValues
          __typename
        }
        __typename
      }
      __typename
    }
    __typename
  }
  distributionMessages {
    low
    middle
    high
    __typename
  }
  __typename
}
fragment RangeValues on StatsRange {
  max
  min
  value
  __typename
}
fragment SectionInfoSection on SectionData {
  name
  stats {
    studentCount
    __typename
  }
  code
  number
  __typename
}
fragment LastUpdatedSection on SectionData {
  lastUpdated
  __typename
}
fragment AppBody on SectionData {
  user {
    type
    __typename
  }
  __typename
}
fragment StudentsMonitoringSection on SectionData {
  code
  ...HelpSection
[...]

PR #45, opened 2017-06-14
making update data interval larger

tnrich:
This hopefully serves as a temporary fix to issue https://github.com/apollographql/apollo-client-devtools/issues/28

### DIFF
--- a/app/components/Inspector/Inspector.js
+++ b/app/components/Inspector/Inspector.js
@@ -65,7 +65,7 @@ export default class Inspector extends React.Component {
     this.updateData();
     this._interval = setInterval(() => {
       this.updateData();
-    }, 2000);
+    }, 5000);
   }
 
   componentWillUnmount() {
--- a/app/components/Inspector/Inspector.js
+++ b/app/components/Inspector/Inspector.js
@@ -65,7 +65,7 @@ export default class Inspector extends React.Component {
     this.updateData();
     this._interval = setInterval(() => {
       this.updateData();
-    }, 500);
+    }, 2000);
   }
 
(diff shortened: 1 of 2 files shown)


rrdelaney:
Hey, thanks so much for submitting this PR! I'm currently working on making the devtools push based (https://github.com/apollographql/apollo-client-devtools/pull/43) but it seems like that's going to go slowly 😕 In the mean time, this looks like a solid fix!

I'll check this out and see if it has any noticeable impact on load times. if not this looks good to go!

stubailo:
Do we have an idea of a good interval for now? We bumped it to 1000, does it really need to go up to 5000? Of course, push based is the right approach long term.

tnrich:
@stubailo , my PR actually bumped it to 5000, but maybe 10000 would be better? It is seemingly a bit arbitrary

stubailo:
`1000` not `10000`. But I think `5000` is super long because then the store will only update every 5 seconds?

tnrich:
Oh, whoops you're right @stubailo

stubailo:
Which panels are you using? Are you using the store panel? What do you use it for

tnrich:
Ahh @stubailo , I don't use the store panel very much because I can view the store in the redux devtools already. I only really look at the watched queries panel (and now watched mutation panel!).

stubailo:
Ah - in that case, this interval shouldn't change anything since this is just the refresh for the store panel. That explains why the refresh didn't make it laggy.

Unless somehow we are polling for store changes all the time.

thomassuckow:
Would probably be better to do something like #51

PR #51, opened 2017-06-26
Sleep for 1 second between data updates

thomassuckow:
Rather than using setInterval which can queue up and run back to back, using setTimeout AFTER the data is updated allows giving the browser 1 second to do other work. This also means that if evens are backed up you have at most 1 update cycle before browser events are handled again instead of the time getting longer and longer as the interval's queue up.

Note this is untested.

### DIFF
--- a/app/components/Inspector/Inspector.js
+++ b/app/components/Inspector/Inspector.js
@@ -63,13 +63,20 @@ export default class Inspector extends React.Component {
   componentDidMount() {
     if (ga) ga('send', 'pageview', 'StoreInspector');
     this.updateData();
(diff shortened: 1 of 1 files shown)


thomassuckow:
Won't sign CLA, closing.

tnrich:
@thomassuckow can I take your code and make the same PR? Any reason in particular why you don't want to sign the CLA? (I'm just curious, I have no affiliation to meteor whatsoever)

thomassuckow:
Take it, test it, improve it.

I won't sign the CLA because Graphql would be deprecated by the time my work lawyers approved it.

PR #53, opened 2017-06-28
Updating polling in both Panel and Inspector to wait for the previous polling result to have finished

tnrich:
Yay! With some more poking around at how the polling was working, I think I've come up with a solution that finally fixes the apollo devtools from crashing my browser tab. 

I realized that polling is not just happening in `Inspector.js`, but also `Panel.js`  (note the rapidly incrementing "polling!" console.log in the gif below:
asdfasdfasd

As well as incrementing the default polling interval in Panel.js to 1000 (similar to the change already implemented for Inspector.js), I've also promisified the polling in both places (I'm not 100% sure that the window communication is async but it is using callbacks so I believe it is) and used the trick @thomassuckow employed in his closed PR of calling a new setInterval only after invoking the polling. 

The end result of this is that now polling in both Panel and Inspector to wait for the previous polling result to have finished! Here's a gif of the polling occurring only after it has finished:

timing

To view the above console logs you can checkout the previous of the 2 commits :)

### DIFF
--- a/app/components/Inspector/Inspector.js
+++ b/app/components/Inspector/Inspector.js
@@ -67,9 +67,7 @@ export default class Inspector extends React.Component {
     if (ga) ga('send', 'pageview', 'StoreInspector');
     this.updateData();
     const updater = () => this._interval = setTimeout(() => {
-      this.updateData().then(() => {
-        updater();
-      })
+      this.updateData().catch(console.error).then(updater)
     }, 1000);
     
     updater();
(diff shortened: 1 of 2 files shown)


thomassuckow:
Note that promisifying will swallow exceptions which could be difficult to locate and the setTimeout wouldn't be reset. You should probably add a catch to log errors:

```javascript
this.updateData().catch(console.error).then(updater);
```

tnrich:
@thomassuckow , nice I took your advice.

tnrich:
@rrdelaney merge it?

rrdelaney:
Waiting till tonight if anyone else wants to review 😄

tnrich:
Sounds good :)

ramyanaga:
Published ~30 minutes ago.

PR #55, opened 2017-07-06
Panel data bridge

ramyanaga:
Update panel to push-based data loading.
Potential Issue: not sure if store is loading all of the data correctly


app/components/Panel.js
```diff
@@ -38,57 +39,49 @@ export default class Panel extends Component {
...
-          this.lastActionId = newLastActionId;
-          this.setState({
-            actionLog: result,
-          });
-        }
-      });
-    }, 100);
-
+     this.interval = setInterval(() => {
+     }, 100);
```
rrdelaney: Did you mean to remove this?

app/components/contentScript.js
```diff
@@ -0,0 +1,3 @@
+chrome.runtime.sendMessage({
```
rrdelaney: Was this file intended to be included?

extension/hook.js
```diff
@@ -1,30 +1,58 @@
...
+      chrome.runtime.sendMessage({ APOLLO_CONNECTED: true}, function() {
+        passedApolloConnected = true;
+      });
+    }
+  }
+
+  if (!!event.data.trimmedObj) {
+    chrome.runtime.sendMessage({ trimmedObj: event.data.trimmedObj });
+  }
+  else {
```
rrdelaney: I don't think you need this `else` statement - a blank return should have the same effect as no return

app/components/Panel.js
```diff
@@ -38,57 +39,49 @@ export default class Panel extends Component {
       automaticallyRunQuery: undefined
     };
 
+    let backgroundPageConnection = chrome.runtime.connect({
+      // sending tabId as name to make connection one-step process
+      name: chrome.devtools.inspectedWindow.tabId.toString()
+    });
+
+    backgroundPageConnection.onMessage.addListener((request, sender) => {
+      request = [request];
```
rrdelaney: Why wrap this in an array and then `.map` over it?

stubailo:
@ramyanaga why is this adding 3000 lines of code? sounds like we included a file we don't need or something

stubailo:
Ohh - package-lock.json from NPM 5. carry on.

app/components/Panel.js
```diff
@@ -124,8 +117,6 @@ export default class Panel extends Component {
             state: window.__APOLLO_CLIENT__.queryManager.getApolloState(),
             dataWithOptimisticResults: window.__APOLLO_CLIENT__.queryManager.getDataWithOptimisticResults(),
           });
-
-          window.__APOLLO_CLIENT__.__actionHookForDevTools(logger);
```
stubailo: I think removing this might make the store panel stop working?

extension/background.js
```diff
@@ -1,6 +1,44 @@
...
+let backgroundPageConnection = undefined;
+let requestLog = {};
+
+chrome.runtime.onConnect.addListener((port) => {
+  tabId = port.name
+  connections[tabId] = port;
+
+  port.onDisconnect.addListener(port => {
+
+    port.onMessage.removeListener(extensionListener);
```
stubailo: I don't think extensionListener exists here.

extension/background.js
```diff
@@ -1,6 +1,44 @@
...
+
+    const tabs = Object.keys(connections);
+    for (let i = 0; i < tabs.length; i++) {
+      if (connections[tabs[i]] == port) {
+        delete connections[tabs[i]];
```
stubailo: I think this entire for loop thingy can just be replaced with

```js
delete connections[tabId];
```

Or am I missing something?

extension/background.js
```diff
@@ -1,6 +1,44 @@
...
+  // not sure if being used
+  if (!apolloConnected && request.APOLLO_CONNECTED) {
     chrome.pageAction.show(sender.tab.id);
+    apolloConnected = true;
+  }
+  if (request.trimmedObj) {
+    try {
+      connections[sender.tab.id].postMessage(request.trimmedObj);
+    }
+    catch(err) {
```
stubailo: When does this error occur? Can we instead add an `if` statement to detect that case? This will catch _all_ errors, not just the one we're specifically trying to handle here.

app/components/Panel.js
ramyanaga: I tried to keep the code as similar to the old version. The comparable section of code in the old version returned the request through an eval in an array and mapped over it, so I ended up doing a similar thing.

stubailo: I think we should make this not an array.

PR #63, opened 2017-07-14
Store tab push-based and active-tab-check data loading

ramyanaga:


app/components/Inspector/Inspector.js
```diff
@@ -38,60 +37,64 @@ export default class Inspector extends React.Component {
       `,
         dataWithOptimistic => {
           let toHighlight = {};
```
stubailo: This line is no longer used

app/components/Inspector/Inspector.js
```diff
@@ -38,60 +37,64 @@ export default class Inspector extends React.Component {
...
   }
 
   componentWillUnmount() {
     this.didMount = false;
     clearInterval(this._interval);
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!nextProps) {
+      return [null, null, null, null];
```
stubailo: What is this doing?

PR #65, opened 2017-07-19
Final Edits of New Data Bridge

ramyanaga:
Final edits to push-based data loading model before merging into master/publishing.


app/components/Inspector/Inspector.js
```diff
@@ -27,68 +25,72 @@ export default class Inspector extends React.Component {
...
 
   componentWillUnmount() {
     clearTimeout(this._interval);
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!nextProps) {
+      return [null, null, null, null];
+    }
+    this.updateDataInStore(nextProps.state.inspector);
```
stubailo: Why not just read this directly from the props, rather than using `setState`?

extension/hook.js
```diff
@@ -53,14 +59,54 @@ window.addEventListener('message', event => {
...
+        type: 'UPDATE_TAB_DATA',
+        inspector: event.data.newStateData.inspector
+      });
+    }
   }
+
   return;
 });
+
+// lines 97 - 108 send data to a tab only when a new tab is opened
```
stubailo: No need to put line numbers - it looks like they are already outdated anyway

stubailo:
Yay! No more page crashing!

PR #67, opened 2017-07-25
[WIP/Need Help] Postmessage issue

ramyanaga:
Fixed postmessage issue (I think?), but I'm confused as to why the solution works, and why my alternate solution doesn't work.

### DIFF
--- a/app/components/Panel.js
+++ b/app/components/Panel.js
@@ -39,15 +39,7 @@ export default class Panel extends Component {
 
       if (logItem.queries) {
         console.log(logItem.queries);
-
         let queries = logItem.queries;
-        /*
-        for (var query in queries) {
-          const variablesObject = logItem.queries[query].variables;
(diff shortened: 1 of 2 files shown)


app/components/Panel.js
```diff
@@ -38,14 +38,16 @@ export default class Panel extends Component {
       let tabData;
 
       if (logItem.queries) {
+        console.log(logItem.queries);
```
stubailo: Both of these lines are unused 😛

extension/hook.js
```diff
@@ -18,12 +18,41 @@ const hookLogger = (logItem) => {
 
   if (!!window.__APOLLO_CLIENT__) {
 
+    let queries = logItem.state.queries;
+    
+    // why does this not work??
```
stubailo: rip

intellix:
This looking good? My developer tools are getting spammed to hell without this fix :(

PR #76, opened 2017-09-14
Handle both networkInterface and link in Explorer

slightlytyler:
Would love to update to v2 of apollo-client but need the dev tools working first. Let me know if this approach is off, down to resubmit if needed.


slightlytyler:
Sure thing, I can take a look this evening @peggyrayzis

jbaxleyiii:
@slightlytyler thank you for this! With the link additions we had to make a change in the rc so I ended up adjusting this. Its on master now so I'm going to close this but thank you again!

PR #85, opened 2017-11-18
Attempt to make Apollo panel creation more reliable

### LINKED ISSUE #29: Devtools not showing up reliably.
Devtools sometimes shows up but mostly doesn't (race condition?)

Webpack: 2.3.3
Extension: 1.0.8.1
Chrome: 57.0.2987.133 (64-bit)
React/Redux/Apollo stack

The extension actually loads: Devtools/Sources shows that the hook.js Content Script has loaded, and indeed window.__APOLLO_DEVTOOLS_GLOBAL_HOOK__ shows the correct version (and of course, the Warning message doesn't appear).

Restarting Chrome sometimes fixes the problem.

Repeat of https://github.com/apollographql/apollo-client-devtools/issues/8

tizmagik:
fixes #29 (hopefully 🤞 )

Using this locally the Apollo panel was created much more reliably. I pattern matched off React devtools [panel creation logic](https://github.com/facebook/react-devtools/blob/master/shells/webextension/src/main.js) as I've never had an issue with the React devtool not showing up 😄

### DIFF
--- a/extension/devtools.js
+++ b/extension/devtools.js
@@ -28,10 +28,10 @@ function createPanel() {
   );
 }
 
-// Attempte to create Apollo panel on navigations as well
+// Attempt to create Apollo panel on navigations as well
 chrome.devtools.network.onNavigated.addListener(createPanel);
 
-// Attempte to create Apollo panel once per second in case
+// Attempt to create Apollo panel once per second in case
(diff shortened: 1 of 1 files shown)


tizmagik:
I’ve signed it! 😁

PR #89, opened 2017-12-14
Removes Broken Firefox Link

stolinski:
https://addons.mozilla.org/en-US/firefox/addon/apollo-client-dev-tools/ currently points to a 404. This is a little confusing to have a broken link here. I didn't see that the Firefox dev tools had been released yet anywhere, so I think it's best to pull the link until it's up and working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Apollo Client Devtools
 ===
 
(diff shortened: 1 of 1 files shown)


jbaxleyiii:
@stolinski we released it but then it was removed, we are working with firefox to figure out why!

Issue #64, opened 2017-07-17
Uncaught DOMException: Failed to execute 'postMessage' on 'Window' #1879 


jbaxleyiii:
From [AC](https://github.com/apollographql/apollo-client/issues/1879)

* could be related to [#1871](https://github.com/apollographql/apollo-client/issues/1871)

I have afterware on my apollo client to handle errors in my app. In order to check the server's response status I have to clone the response. It was working fine until suddenly (chrome update?) I started getting this message:
`DOMException: Failed to execute 'postMessage' on 'Window': Error: Network request failed with status 401 - "Unauthorized" could not be cloned.`
(the error is the response from the server).

To replicate, create an afterware that clones the response from the server.

Thanks!

ramyanaga:
Are you trying to post objects that are not Json-serializable? If so, this is the same issue as #58

stubailo:
Ah! Maybe this error has to do with errors being displayed in the dev tools?

tnrich:
Hitting this error as well. Does anyone have an idea why having an afterware with `response.clone() `  would be causing this to occur? I don't follow the logic there. Thanks!


```
VM146910:13 Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Window': function xhrAdapter(config) {
  return new Promise(function dispatchXhrRequest(resolve, reject) {
    var reque...<omitted>...
} could not be cloned.
    at ApolloClient.hookLogger [as devToolsHookCb] (<anonymous>:13:12)
    at Object.dispatch (http://localhost:4000/static/js/bundle.js:133366:31)
    at dispatch (<anonymous>:2:1620)
...
```

ramyanaga:
@tnrich The breaking is in the hookLogger function defined in hook.js; its because chrome's postMessage function can't handle non-serializable objects. (I think)

tnrich:
@ramyanaga , thanks for the info!

Just to be clear, you're referring to this line in particular
```js
  const newStateData = {
      queries: logItem.state.queries,
      mutations: logItem.state.mutations,
      inspector: logItem.dataWithOptimisticResults
    }
    window.postMessage({ newStateData }, '*');
```
https://github.com/apollographql/apollo-client-devtools/blob/master/extension/hook.js#L27

That's what looked like the culprit might be to me as well. Is there something simple we could do to provide better feedback to the user? Perhaps something like:
```js
	try {
		JSON.stringify({newStateData})
	} catch (e) {
[...]

Issue #58, opened 2017-07-09
postMessage fails when query information contains non-serializable values?


stubailo:
https://github.com/apollographql/apollo-client/issues/1871

Looks like it has trouble with some objects like `Date`  types which are still OK when JSON serialized and sent to a GraphQL server.

tnrich:
Anything I can do to help fix this? Running into this issue now and am not sure how to debug it, but would love to have the apollo devtools back up and running.

ieow:
Any update on this issue?
I am having same messages for every graphql error 
`DOMException: Failed to execute 'postMessage' on 'Window': Error: GraphQL error:`

It keep printing every time I try query graphql after the error.

rohmanhm:
Weird error, but I got the same issue here.

PR #91, opened 2018-01-03
Fix #75 #64 #58

### LINKED ISSUE #75: DOMException: Failed to execute 'postMessage' on 'Window'
There are two other `postMessage` related issues logged here (#64, #58), but they both seem related to passing non-serialized objects. 

In my case, a simple mutation as described in the "[Apollo User Auth w/ Email](https://www.graph.cool/docs/tutorials/react-apollo-email-oopheesaj9)" guide will throw this error if executed more than once. 

The error isn't thrown with the Apollo Dev Tools disabled.

pSchaub:
This will create a full deep clone of the object before passing it to `postMessage`. This will fix #75, #64 and #58.

Technical information:
We use the `JSON.parse(JSON.stringify(obj))` trick to do a deep clone of the object.
It's the fastest way to deep clone an object, see https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript

### DIFF
--- a/extension/hook.js
+++ b/extension/hook.js
@@ -19,7 +19,7 @@ const hookLogger = (logItem) => {
     }
     
     try {
-      window.postMessage({ newStateData }, '*');  
+      window.postMessage(JSON.parse(JSON.stringify({ newStateData })), '*');  
     }
     catch(err) {
       console.log(err);


pSchaub:
@meteor-bot done ;-)

gdi2290:
can we get this merged

PR #97, opened 2018-01-31
Rebuild

jbaxleyiii:
Thanks to the help from @akryum, this now is way more stable! And is a starting point for a whole lot more work to come!


Akryum:
Thanks for the shout out! Does this PR means we will get a Firefox addon soon? :smile:
(Don't forget to use `yarn pack` or `npm pack` to create a zip of the sources, so you can upload it to addons.mozilla.org while publishing each version.)

PR #98, opened 2018-02-01
Remove broken link to Firefox dev tools

wesbos:
As far as I can tell Apollo dev tools doesn't exist for Firefox, so I'm removing the link


jbaxleyiii:
@wesbos its waiting on a review 😢 

but yeah we should remove it until its ready. Thanks!

PR #125, opened 2018-05-22
Fixed issue when query.queryString doesn't exist anymore

### LINKED ISSUE #119: Crashes on clicking Queries
[Dev](url) tools is crashing whenever I click on Queries. Similar to what is happening in #107 

Mac OS 10.13.3
Chrome 66.0.3359.139
screen shot 2018-05-15 at 5 56 22 pm

Negan1911:
This fixes an issue encountered on react-apollo 2 which closes #119

### DIFF
--- a/src/devtools/components/Mutations/Mutations.js
+++ b/src/devtools/components/Mutations/Mutations.js
@@ -12,7 +12,9 @@ import Warning from "../Images/Warning";
 import "../WatchedQueries/WatchedQueries.less";
 
 const mutationLabel = (mutationId, mutation) => {
-  const mutationName = getOperationName(parse(mutation.mutationString));
+  const mutationName = getOperationName(
+    parse(mutation.mutationString || mutation.document.loc.source.body),
+  );
   if (mutationName === null) {
     return mutationId;
   }
@@ -180,6 +182,9 @@ class WatchedMutation extends React.Component {
       mutation.metadata &&
       mutation.metadata.reactComponent &&
       mutation.metadata.reactComponent.displayName;
+
+    const mutationString =
+      mutation.mutationString || mutation.document.loc.source.body;
     return (
       <div className={classnames("main", { loading: mutation.loading })}>
         <div className="panel-title">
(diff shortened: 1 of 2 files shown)


Negan1911:
Quick question, maybe instead of replacing would make sense to check if exists first? like `query.queryString || query.document.loc.source.body `

umlx5h:
```
diff --git a/src/devtools/components/WatchedQueries/WatchedQueries.js b/src/devtools/components/WatchedQueries/WatchedQueries.js
index 8ecf768..ba2a6df 100755
--- a/src/devtools/components/WatchedQueries/WatchedQueries.js
+++ b/src/devtools/components/WatchedQueries/WatchedQueries.js
@@ -52,6 +52,10 @@ class WatchedQueries extends React.Component {
     const hasError =
       query.networkError ||
       (query.graphQLErrors && query.graphQLErrors.length > 0);
+
+    if (!query.queryString) {
+      query["queryString"] = query.document.loc.source.body;
+    }
     return (
       <li
         key={id}
```

it's might be a more elegant way to fix.

Negan1911:
I dont like to mutate an object but the final words are from a maintainer, let's wait. If someone thinks that that's more elegant I'll update my PR

pthrasher:
@Negan1911 I'm not a contributor, but I like your suggestion. I also don't think it's a good idea to modify the object. Better to move forward with the correct API while still allowing support for the old one.

Would be nice to get this merged in.

Negan1911:
Thanks, also i think that we can get ahead of time and just do the same for mutations, i'll update my PR to reflect that, the only thing that concerns me is the support of `query.document.loc.source.body ` on old platforms

Negan1911:
@peggyrayzis Saw your commit, thanks! quick question, i pulled your commit and did the same for mutations, do you think that is worth to push it?

peggyrayzis:
@Negan1911 Good catch! 🎉 Can you please make the fix for mutations as well? It looks like the mutation tab is busted right now - I can't click show in GraphiQL to run the mutation. Once you push that fix, I'll merge.

peggyrayzis:
Also once you make the fix, could you please post an animated gif confirming the change so I don't have to rebundle the extension again to test manually? 😀 @Negan1911

Negan1911:
Will do!

Negan1911:
@peggyrayzis Actually, can't link it here because of the file size, but you can see it here: https://i.imgur.com/NHCWsOI.gifv

peggyrayzis:
Awesome! Thanks @Negan1911 :shipit:

PR #133, opened 2018-07-11
Fix issue where "Run in GraphiQL" does not including fragments

henryqdineen:
This is a fix to the changes in #125.

This fixes an issue that my team is running in to that is blocking us from using the dev tools and especially the `"Run in GraphiQL"` feature.

The `query.document.loc.source.body` will only include the query and not any of the fragments. My solution was the print the entire AST to a string. 

Let me know if there's anything else you need from me. Thanks!

Notice in the screenshot that the fragments are missing from the document that is loaded into GraphiQL. 
screenshot 2018-07-12 18 15 15

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 ## vNext
 
 * Change cursor in dark theme to white ([#131](https://github.com/apollographql/apollo-client-devtools/pull/131))
+* Fix issue where "Run in GraphiQL" does not include fragments ([#133](https://github.com/apollographql/apollo-client-devtools/pull/133))
 
 ## 2.1.3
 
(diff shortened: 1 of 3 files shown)


henryqdineen:
out of curiosity does anybody know the history of the `query.queryString` usage is? Is it to support an older version of apollo-client?

henryqdineen:
Is anyone available to look into this? 

Our app makes heavy use of fragments (was converted from a relay modern app with hundreds of "fragment containers") and this bug has made it a challenging to use the apollo dev tools.

henryqdineen:
Just checking in on this. I'm surprised people don't run into this more often when using fragments so it makes me wonder if we are doing something wrong our side?

piotrd:
@henryqdineen I've been annoyed by this bug for a while, finally I built a version of extension from your PR and I can say it works as advertised. Thanks!

PS. Indeed it's odd that the interest in this issue is low, but I hope someone with eventually merge it.

henryqdineen:
great! thanks @hwillson

henryqdineen:
@hwillson any idea when this would get into a release? Was planning to do an interactive demo to some of our devs and wanted them to be able to follow along.

PR #138, opened 2018-09-14
Upgrade to babel-7, webpack-4, and more

ianks:
This PR performs upgrades of a few crucial components:

1. Upgrade to babel 7
2. Upgrade to webpack 4
3. A smorgasbord or other dependencies.
4. Use yarn instead of npm

In summary, all dependencies have are now up to date besides `react-native-web`. I did not attempt to upgrade that because it seems like there were many breaking changes that I personally have no experience with.

I tested it out with `yarn firefox` and it seems to be working properly in my apollo app. When running `yarn dev`, however, it could not connect to an Apollo client. _Am I doing it wrong?_

I would really love to get some eyes on this so I can fix any issues :smile: 

To test it out:

```
git clone https://github.com/ianks/apollo-client-devtools 
yarn
yarn build
yarn firefox
# browse to site with apollo client with devtools enabled and open up the developer console, find Apollo tab
```

Cheers!


hwillson:
Thanks for working on this @ianks! Unfortunately, the changes listed here aren't something we can accept in the current format. Your babel and webpack dep updates are great, but there are other changes in this PR that head in a direction we don't want to head in. For example, we don't want to rename the extension of JSX files, we don't want to switch to using yarn, etc. If you're interested in creating a new PR that just includes the babel and webpack updates (and have validated that everything is working), we'll definitely take a look. Thanks again for your time!

Issue #104, opened 2018-02-09
Cannot read property 'queryStore' of undefined, cannot convert null or undefined to object, unsafe-eval is not allowed source of script in the following CSP


hinok:
```
npm ls --depth=0 | grep apollo
├── apollo-cache-inmemory@1.1.5
├── apollo-client@2.2.1
├── apollo-fetch@0.7.0
├── apollo-link-batch-http@1.0.4
├── react-apollo@2.0.4
```

Hey,
In our project apollo-client-devtools stopped working.
In the console inside the app, there is a reported error:

![zrzut ekranu 2018-02-09 o 16 17 08](https://user-images.githubusercontent.com/1313605/36034622-eb17755e-0db4-11e8-9c86-7193765f49fc.png)

which is:

![zrzut ekranu 2018-02-09 o 16 17 32](https://user-images.githubusercontent.com/1313605/36034630-efcab80e-0db4-11e8-838f-87a173898aa3.png)

Also when I try to debug apollo devtools, in their console are reported errors below (here devtools has disappeared):

![zrzut ekranu 2018-02-09 o 16 16 48](https://user-images.githubusercontent.com/1313605/36034655-02e25226-0db5-11e8-85d1-c734995e45f2.png)

![zrzut ekranu 2018-02-09 o 16 16 31](https://user-images.githubusercontent.com/1313605/36034668-0bd75048-0db5-11e8-8178-75ccc1883178.png)

Is this something that we should fix on the our side?
Many thanks.

steve-a-jones:
Getting this exact same error - happens when I click on the cache tab. 

`Chrome: Version 64.0.3282.140`

ostrebler:
Same here.

BrianSandholdt:
I experience the same thing. Could this be an issue with latest update of Chrome?

mfeeney:
Same here. 

`Version 64.0.3282.167 (Official Build) (64-bit)`

ar-sp:
I had the same error (Cannot read property 'queryStore' of undefined), but it was on an new project which I hadn't set up properly. My `InMemoryCache.fragmentMatcher` wasn't set up right, and when I fixed it the devtools problem was resolved.

Not sure if it will help anyone else here or not, but might be useful for future google hits :)

alex88:
Same here with regular default installation that uses `InMemoryCache` with default settings

sundeer:
Same here.
[...]

PR #139, opened 2018-09-17
Set initial state to empty if queryManager does not exist

adampetrie:
This is an attempt to fix https://github.com/apollographql/apollo-client-devtools/issues/104.

When the dev tools load and the client has yet to run a query, the `panel:ready` event handler blindly tries to get the `queryStore` and the `mutationStore` from the `queryManager` even though that property is not set on the client object until a query or mutation gets run.

The solution was to follow the same logic that gets used in [initQueryManager()](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/ApolloClient.ts#L402) and set these values to be empty objects if the `queryManager` does not exist.

The logic I referenced above is used in all calls to the `devToolsHookCb` function in the client which is handled by the `logger` arrow function in the modified file which leads me to believe that this is safe change to make.

### DIFF
--- a/src/backend/broadcastQueries.js
+++ b/src/backend/broadcastQueries.js
@@ -17,8 +17,12 @@ export const initBroadCastEvents = (hook, bridge) => {
   bridge.on("panel:ready", () => {
     const client = hook.ApolloClient;
     const initial = {
-      queries: client.queryManager.queryStore.getStore(),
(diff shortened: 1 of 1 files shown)


jamesjryan:
Can we get this merged? Seems to be affecting basically everyone trying to view the Cache tab.

Issue #128, opened 2018-06-06
Broken Firefox Addon Link


dan003400:
The link to the Firefox DevTools addon is broken:

https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/

dan003400:
Me either so I was assuming it was taken down?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

CarloPalinckx:
Is there any news on this? :)

DSM3DIA:
Would be nice to have this. Trying to switch over completely from chrome...chrome is sucking up 17gigs of ram on my 32gig workstation

ngmichael:
I would appreciate having the devtools for Firefox as well.

saostad:
+1

nathanrodman:
+1 for firefox devtools

cyberbizness:
Toc Toc... Quelqu'un s'en occupe?

therealparmesh:
+1

Extarys:
We do need a Chrome alternative.

dvdbot:
@Extarys This is about the Firefox version - please open a new issue for the Chrome Version

Extarys:
@dvdbot it is concerning Firefox, as I said, we need something else than just Chrome. :smile:

darylf:
This is a duplicate of #114. Note, there is an open request which is over a year old requesting Firefox support (see #73)

Although I haven't tried it, @eschaefer has a temporary workaround that seems interesting: https://github.com/apollographql/apollo-client-devtools/issues/114#issuecomment-388805335

ghirlekar:
With all the privacy issues surrounding Chrome the last few days, a lot of us would love to switch over to Firefox. Apollo dev tools is the last extension I need in order to do that.

Anyone knows why the plugin was pulled from the Firefox extension store?

Dieff:
For anyone seeing this now, you can manually build the this addon and use it in Firefox.

1. Clone the repo
2. Run `npm install && npm run build`
3. Navigate to `about:debugging` in Firefox
4. Click "Load Temporary Add-on"
5. Navigate to /repo_root/sheels/webextension and load that folder as an addon

I received a warning during the above steps, but the addon itself appears to be working fine.
Maybe someone would be interested in uploading an unofficial version to the Firefox extension site?

Dieff:
[...]

Issue #114, opened 2018-03-21
Firefox link is broken


Pajn:
The firefox addon link gives me a 404 and I can not find the extension on the Firefox add on site manually.

vaneek:
@jbaxleyiii is Firefox addon supported or is development and testing occurring on the master branch?

iamtchelo:
Still broken! :/

eschaefer:
Until this is resolved, you can of course build the extension locally from this repo and install it on Firefox

```
git clone https://github.com/apollographql/apollo-client-devtools.git
npm i
npm run build
```
Open "about:debugging" in Firefox, click "Load Temporary Add-on" and select the `manifest.json` in `shells/webextension`.

Refresh the page you need to debug, and voila: 
<img width="1367" alt="screen shot 2018-05-14 at 14 44 41" src="https://user-images.githubusercontent.com/547148/39998365-28ccadc2-5786-11e8-8e45-53cf9f72570f.png">

RobertB4:
Still no update? May as well switch to chrome until the firefox dev tools are working again.

rkstar:
i've just gone back to firefox and i'm enjoying the experience.... really need this extension tho.

joejordan:
So is Firefox supported or not? This issue has been open for 8 months(!) and the readme is still pointing to a deleted add-on page (https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/)

hwillson:
Please see https://github.com/apollographql/apollo-client-devtools/issues/73#issuecomment-444497422. In a nutshell - devtools supports FF, but has been pulled by FF. We're working on a fix, and will track all progress in #73. Thanks!

Issue #73, opened 2017-08-21
Firefox support


avrelaun:
Can we have official firefox support for this addon ?

The build from https://github.com/apollographql/apollo-client-devtools/pull/41 works for now, but it's not durable.

tizmagik:
🤞 https://twitter.com/peggyrayzis/status/933741807942488071 😁

fandy:
Any update on this @peggyrayzis? I would love to have this moved to FF since with Quantum it is now my browser of choice.

peggyrayzis:
@fandy I published devtools to the Firefox addons store three weeks ago. Everything was working fine for a week until it recently got taken down. I haven't had the time to look into it and won't be able to get to it until after the holidays, unfortunately.

maticrivo:
any news on this?

EthanML:
This would be awesome to see happen again - though if it's not likely anytime soon perhaps the link to an FF version should be removed from the README for now.

frankdugan3:
I haven't had any luck installing from the latest release files either. Firefox simply throws an error saying that the add-on appears to be corrupt. Are there any special steps to installing it from the release files?

haikyuu:
maybe https://github.com/mozilla/addons/issues/450 is related

seangransee:
Agreed that the link should be removed from the README until this is fixed.

Extarys:
I want it :( RIP

japrogramer:
+1

relhtml:
What's the latest on Apollo Developer Tools support on the latest Firefox?

takticcode:
No Firefox support yet?

hwillson:
Firefox is supported, but currently broken. FF has pulled devtools due to its use current use of `eval`. We're using `eval` in a safe way, but even if we plead our case and can get FF to agree, I'd like to move away from using `eval` as much as possible (to prevent this from happening again in the future). I'm working on a solution to this, and hope to have this all resolved very shortly. Thanks all!

hwillson:
Devtools 2.1.6 should now be available to Firefox users: 
https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/

Let me know if you encounter any issues. Thanks!

drdaeman:
The link says "page not found" for me. Was the addon pulled back from the directory?

jacobmischka:
I imagine it was, it was kinda buggy. I'm guessing it'll be back as soon as those kinks get ironed out!

[...]

PR #140, opened 2018-09-18
Remove Broken Firefox Addon Link

darylf:
As mentioned in Issue #114 and #128, the link to the Firefox Addon Repository is broken. It appears the plugin has been pulled down. This is confusing to new contributors who discover this code base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Apollo Client Devtools
 
-[Download for Chrome](https://chrome.google.com/webstore/detail/apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm) | [Download for Firefox](https://addons.mozilla.org/en-US/firefox/addon/apollo-developer-tools/)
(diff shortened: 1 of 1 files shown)


robinmetral:
IMHO removing the link altogether isn't the best option.

I landed on this repo already looking for the Firefox addon (suggested in a Wes Bos course), so if the mention was removed from the README, I wouldn't know the addon is down and would just keep looking for it.

I think a better patch would be [suggesting workarounds](https://github.com/apollographql/apollo-client-devtools/issues/114#issuecomment-388805335) in the README, and/or letting people know when the Firefox addon will be back online!

darylf:
I don't necessarily disagree with you, and as a Firefox user I'm naturally disappointed by the lack of acknowledgement by the team; however, if Firefox isn't officially supported, I don't believe it should be mentioned or linked to.

hwillson:
Please see https://github.com/apollographql/apollo-client-devtools/issues/73#issuecomment-444497422. In a nutshell - devtools supports FF, but has been pulled by FF. We're working on a fix, and will track all progress in #73. Thanks for taking the initiative to address this!

PR #148, opened 2018-10-24
fix(WatchedQueries): Fix query name display

tknickman:
### TL;DR
Watched Queries panel in the apollo-client-devtools does not display the correct title for queries stored in the same file. 

# Description
I noticed this bug the other day when refactoring our GQL query names. We have a few components that use multiple queries. We store these queries in one file, for example:
```
componentName/
    index.js // component code
    componentName.graphql // graphql queries used by componentName
```
`componentName.graphql` in this case looks like:
```
query exampleGQLQuery1($exampleId: ID!, $trendBy: TrendTimePeriod) {
    ...
}

query exampleGQLQuery2($exampleId: ID!) {
   ...
}
```

Inside `componentName/index.js` we import those queries via:
`import { exampleGQLQuery1, exampleGQLQuery2 } from './componentName.graphql';`

---

I wanted to check out these queries using the `apollo-client-devtools` (great tool btw! 😄), and I noticed that the Watched Queries tab could did correctly interpret the query names for multiple queries within the same file. The query string and variables were all correct, but the title (both in the query sidebar, and in the query panel after it has been selected) was wrong (it just uses the name of the first query in the file). 

Here are some examples (with some of our private schema data redacted):

### Watched Queries Zero State
You can see the two queries listed in the sidebar with the same query name - when in reality, these are two distinct queries with different names stored in the same file
zeroStateBroken

### Watched Queries: First Query Selected.
After selected the first query in the sidebar, you can see the query string and variables are correct (because this is the first query in the file, the title matches)
query1Broken

### Watched Queries: Second Query Selected.
After selected the second query in the sidebar, you can see the query string and variables are correct, but the but the title in the sidebar, and in the panel is wrong (this is because the first query in the file is being used to generate all of the query titles.
query2Broken

### Watched Queries Zero State
You can see the two queries listed in the sidebar with their correct query names
zeroStateBroken

### Watched Queries: First Query Selected.
After selected the first query in the sidebar, you can see that the titles, query string and variables are correct.
query1Broken

### Watched Queries: Second Query Selected.
After selected the second query in the sidebar, you can see that the titles, query string and variables are correct.
query2Broken

Let me know what you think! Happy to provide more examples etc. as needed 😄

### DIFF
--- a/src/devtools/components/WatchedQueries/WatchedQueries.js
+++ b/src/devtools/components/WatchedQueries/WatchedQueries.js
@@ -13,6 +13,13 @@ import Warning from "../Images/Warning";
 import "./WatchedQueries.less";
 
 const queryLabel = (queryId, query) => {
+  const queryDefinitions =
+    query && query.document && query.document.definitions;
+
+  if (queryDefinitions.length === 1) {
+    return queryDefinitions[0].name.value;
+  }
+
   const queryName = getOperationName(
     parse(query.queryString || query.document.loc.source.body),
   );


tknickman:
Looks like this was fixed (in probably a better way ) in https://github.com/apollographql/apollo-client-devtools/pull/149

Closing.

Issue #160, opened 2018-12-21
Publishing as npm module


Gongreg:
Hello, I would like to know, would it be possible to publish apollo-dev-tools as a npm module so it could be included as a dependency in [react-native-debugger](https://github.com/jhen0409/react-native-debugger)? We need to use some code (backend code of apollo devtools) inside the debugger itself, to support apollo dev tools for react-native.

A slightly offtopic question, but maybe you know what exactly the UUID is used here for? https://github.com/apollographql/apollo-client-devtools/blob/dc0d97f42e1223ea80e464bdc7aa518006a23c07/src/backend/checkVersions.js 

I am having issues with this part of code since it uses localStorage, but localStorage is not available inside webworkers (react native runs in webworker).

Also here is PR for implementing apollo dev tools for react native.  https://github.com/jhen0409/react-native-debugger/pull/298

danilobuerger:
@hwillson could you take a look at this please so we can finally get react-native dev tool support working? It should be easy to publish this?

Gongreg:
It is holidays season now :) Lets hope to release it in beginning of January :)

hwillson:
Thanks for heading up this initiative @Gongreg! We should be able to do this. The Apollo team is a bit light this week due to the holidays, but I've flagged this for discussion when everyone gets back next week.

Gongreg:
Hey @hwillson,
Maybe you have any news?

Gongreg:
Hey @hwillson, I would really like to move this forward.
I saw a lot of positive feedback from community, the sooner we release it the better :)

Gongreg:
Hello @hwillson,
Writing as a reminder. Is there anything I can do to help?

hwillson:
Hi @Gongreg - we're transitioning repo management responsibilities a bit internally. @cheapsteak and @justinanastos are going to help get outstanding devtools issues under control, PR's reviewed/merged, and work on improving devtools as a whole. Both this request and your great PR https://github.com/apollographql/apollo-client-devtools/pull/165 will be reviewed shortly. Thanks!

Gongreg:
Hey, @cheapsteak
Regarding uuid everything is already solved in the PR I’ve created. It is referenced inside this issue.

Gongreg:
Since it is published I am closing the issue, thanks!

PR #165, opened 2019-01-09
Adding code necessary for React Native support

Gongreg:
In React Native debugger we create our own Bridge that we want to use with apollo dev tools.
I've updated the code to allow to setup custom bridge.

The main question I have is uuid in check versions is used for? This code is not easy to implement in React Native, since it doesn't have localStorage but instead asyncStorage which is, well, async.

Of course for RN we also need to solve #160.

### DIFF
--- a/shells/webextension/src/devtools-background.js
+++ b/shells/webextension/src/devtools-background.js
@@ -34,26 +34,29 @@ function createPanel() {
 
   panelLoaded = false;
   panelShown = false;
-  chrome.devtools.inspectedWindow.eval(`true`, function(result, isException) {
-    // XXX how should we better handle this error?
-    if (isException) console.warn(isException);
+  chrome.devtools.inspectedWindow.eval(
+    `!!(window.__APOLLO_DEVTOOLS_GLOBAL_HOOK__.ApolloClient || window.__APOLLO_DEVTOOLS_SHOULD_DISPLAY_PANEL__);`,
+    function(result, isException) {
+      // XXX how should we better handle this error?
+      if (isException) console.warn(isException);
 
-    // already created or no ApolloClient
-    if (!result || panelCreated) return;
+      // already created or no ApolloClient
+      if (!result || panelCreated) return;
 
-    // clear watcher
-    if (loadCheckInterval) clearInterval(loadCheckInterval);
-    panelCreated = true;
-    chrome.devtools.panels.create(
-      "Apollo",
-      "./imgs/logo_devtools.png",
-      "devtools.html",
-      panel => {
-        panel.onShown.addListener(onPanelShown);
-        panel.onHidden.addListener(onPanelHidden);
-      },
-    );
-  });
+      // clear watcher
+      if (loadCheckInterval) clearInterval(loadCheckInterval);
+      panelCreated = true;
+      chrome.devtools.panels.create(
+        "Apollo",
+        "./imgs/logo_devtools.png",
+        "devtools.html",
+        panel => {
+          panel.onShown.addListener(onPanelShown);
+          panel.onHidden.addListener(onPanelHidden);
+        },
+      );
+    },
+  );
 }
 
 // Attempt to create Apollo panel on navigations as well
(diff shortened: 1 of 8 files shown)


src/backend/checkVersions.js
```diff
@@ -1,11 +1,6 @@
...
 
-let uuid = localStorage.getItem("uuid");
-
-if (!uuid) {
-  uuid = genUuid();
-  localStorage.setItem("uuid", uuid);
-}
+let uuid = genUuid();
 
 export const checkVersions = (hook, bridge) => {
```
benjamn: > The main question I have is uuid in check versions is used for? This code is not easy to implement in React Native, since it doesn't have localStorage but instead asyncStorage which is, well, async.

It seems like `checkVersions` ultimately does an async `fetch` call, so perhaps we could store the `uuid` in an abstract async local storage (backed by sync `localStorage` in browses but async storage in RN), and take advantage of the already-async behavior of `checkVersions`?

Gongreg: Great idea @benjamn, I will implement it today!

Gongreg:
Hey, @benjamn,
I've updated the code. I see that for devtools version the version is taken out of manifest.json.

I am not sure it is the right thing to do to try to get it the same way in react native.
So what I assume is that the version will be in package.json (since it has to have a version). Is that okay?

src/backend/index.js
```diff
@@ -4,10 +4,14 @@ import { initBroadCastEvents } from "./broadcastQueries";
...
 import { checkVersions } from "./checkVersions";
 
-// hook should have been injected before this executes.
-const hook = window.__APOLLO_DEVTOOLS_GLOBAL_HOOK__;
+let hook;
 let bridge;
 let connected;
+let storage;
+
+export const sendBridgeReady = () => {
```
cheapsteak: Would you mind adding a comment here to explain why this function exists and is exported instead of inlined into the call site 🙏? I assume this is so the react native debugger can import and call this :)

shells/webextension/src/devtools-background.js
```diff
@@ -35,7 +35,7 @@ function createPanel() {
   panelLoaded = false;
   panelShown = false;
   chrome.devtools.inspectedWindow.eval(
-    `!!(window.__APOLLO_DEVTOOLS_GLOBAL_HOOK__.ApolloClient);`,
+    `!!(window.__APOLLO_DEVTOOLS_GLOBAL_HOOK__.ApolloClient || window.__APOLLO_DEVTOOLS_SHOULD_DISPLAY_PANEL__);`,
```
cheapsteak: Would you mind adding a comment here as well to explain what  `__APOLLO_DEVTOOLS_SHOULD_DISPLAY_PANEL__` is?

Gongreg: Hey, since there is no Apollo client in the window context of react native debugger (the react native code with apollo is ran in a separate worker), we can't check for __APOLLO_DEVTOOLS_GLOBAL_HOOK__. To make it explicit I add a separate flag.

Gongreg:
Hey @cheapsteak, I've added the comments :)

Gongreg:
Yea, that would be great.
Then I will do final checks with the react native debugger itself and it should be good to go :)

(I've worked on two projects, while doing changes both in checked out projects and in node_modules, in two different machines, so there is a big chance I might have missed something :D)

Can you notify me when the new npm version is published?

cheapsteak:
That sounds great :D 
Yup I can do that

cheapsteak:
@Gongreg published under 2.1.7-alpha.0 for now for testing 🙏

Gongreg:
hey @cheapsteak, seems like the release didn't go well. I still get the code without the PR in 2.1.7-alpha.0 .

https://registry.npmjs.org/apollo-client-devtools/-/apollo-client-devtools-2.1.7-alpha.0.tgz

cheapsteak:
@Gongreg apologies, my bad, just published alpha.1

Gongreg:
@cheapsteak Just tested things out. It works! :) Going to try to get the react-native-debugger pr to be merged as soon as possible :)

justinanastos:
https://github.com/apollographql/apollo-client-devtools/issues/169

cheapsteak:
Published alpha.2 that includes the dist folder
https://unpkg.com/apollo-client-devtools@2.1.7-alpha.2/shells/webextension/dist/

PR #166, opened 2019-01-18
AC-11 Support AC local state

cheapsteak:
99.9% of this was done by Hugh, really appreciated the extremely thorough comments
I just moved a few lines around and double checked that it's working against our full stack tutorial (which uses AC local state) and pupstagram (which uses link-state)

I did find that running link-state mutations doesn't work (in the case of pupstagram, running the `toggleLikedPhoto` mutation results in a `getCacheKey is not a function` error), but it's broken the same way in the currently published version of devtools as well, so it's likely that it hasn't worked in the past

I'm not sure if there's much else worth saying here. The meat of the changes is in `links.js`, and Hugh's comments do a great job of explaining what and why

### DIFF
--- a/src/backend/links.js
+++ b/src/backend/links.js
@@ -109,7 +109,7 @@ export const initLinkEvents = (hook, bridge) => {
       // is being used.
 
       const supportsApolloClientLocalState =
-        typeof apolloClient.getTypeDefs === "function";
+        typeof apolloClient.typeDefs !== "undefined";
 
       if (!supportsApolloClientLocalState) {
         // Supports `apollo-link-state`.
@@ -135,7 +135,7 @@ export const initLinkEvents = (hook, bridge) => {
 
       // Supports Apollo Client local state.
 
-      const typeDefs = apolloClient.getTypeDefs();
+      const typeDefs = apolloClient.typeDefs;
 
       // When using `apollo-link-state`, client supplied typeDefs (for a
       // local only schema) are extracted by re-using the same Apollo Link
--- a/src/backend/links.js
+++ b/src/backend/links.js
@@ -72,33 +72,6 @@ const cacheLink = fetchPolicy =>
     return forward(operation);
   });
 
-const subscribeWithLegacyLinkState = ({
-  key,
-  cache,
-  fetchPolicy,
-  userLink,
-  operationName,
-  subscriptionHandlers,
-}) => {
-  const context = { __devtools_key__: key, cache };
-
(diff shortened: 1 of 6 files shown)


hwillson:
@cheapsteak Let's leave this un-merged for now, just until we know the official launch date of the Apollo Client local state changes. Thanks again!

src/backend/links.js
```diff
@@ -77,32 +73,128 @@ const cacheLink = fetchPolicy =>
...
+      //
+      // The `apollo-link-state` approach uses a custom link chain to parse
+      // and execute queries, whereas the Apollo Client local state approach
+      // uses Apollo Client directly. To decide which approach to use
+      // below, we'll check to see if typeDefs have been set on the
+      // ApolloClient instance, as if so, this means Apollo Client local state
+      // is being used.
+
+      const supportsApolloClientLocalState =
+        typeof apolloClient.getTypeDefs === "function";
```
benjamn: @hwillson Just a note that https://github.com/apollographql/apollo-client/pull/4361/commits/d3931b8347c76c961ae24da68ef108576e67ff0d may have broken this logic (which is fine as long as we make the logic work again before we ship `apollo-client@2.5.0`).

benjamn: With this commit, the new way to get the client `typeDefs` will be `apolloClient.typeDefs`: https://github.com/apollographql/apollo-client/pull/4361/commits/adfd9c008c1a9afab09fdda70f0b121287ed6eb7

justinanastos: Thank you @benjamn ! @hwillson had warned me that `getTypeDefs` was going to be replaced so I've been ready for the breaking change 😃 If this will be released in the latest alpha I'll update the DevTools PR and merge.

src/backend/links.js
```diff
@@ -77,32 +73,126 @@ const cacheLink = fetchPolicy =>
...
+          operationName,
+          context,
+        });
+
+        operationExecution$.subscribe(subscriptionHandlers);
+        return;
+      }
+
+      // Supports Apollo Client local state.
+      const typeDefs = apolloClient.localState.typeDefs;
```
benjamn:
```suggestion
      const typeDefs = apolloClient.typeDefs;
```
This changed again, for reasons explained in this commit message: https://github.com/apollographql/apollo-client/pull/4361/commits/adfd9c008c1a9afab09fdda70f0b121287ed6eb7

justinanastos: Thanks @benjamn !

PR #167, opened 2019-01-18
Prepare to publish to npm (add README for npm and add version to package.json)

cheapsteak:
In preparation for https://github.com/apollographql/apollo-client-devtools/issues/160

The `README` file is the one that's shown for npm, hopefully this would mitigate confusion if anyone finds this package via npm search

### DIFF
--- a/README
+++ b/README
@@ -1,0 +1,5 @@
+# Apollo Client Devtools
+
+This package is published so that some modules can be imported by other devtools.
+
+Please see our [Github Repo](https://github.com/apollographql/apollo-client-devtools) for more information.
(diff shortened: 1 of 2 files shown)


cheapsteak:
@Gongreg good point, I've set it to 2.1.6 now
I wonder if there's a good way to ensure those are synced?

README
```diff
@@ -0,0 +1,5 @@
+# Apollo Client Devtools
+
+This package is published so that some modules can be imported by other devtools.
```
hwillson: Maybe a slight wording adjustment here (so we're explicit):

```suggestion
This package is published so that some of its modules can be imported by other devtools.
```

PR #178, opened 2019-02-14
Make prettier setup more developer-friendly

justinanastos:
- Remove git-staged in favor of [`precise-commits`](prettier.io/docs/en/precommit.html#option-4-precise-commits-https-githubcom-jameshenry-precise-commits)
- Pre-prettify everything so there are no inadvertent prettier-related commits

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+package.json
+package-lock.json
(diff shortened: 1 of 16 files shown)


justinanastos:
> Nice! iirc lint-staged commits the whole file's changes and caused a few accidental commits of parts of a file I didn't intend to commit (console logs, unrelated changes etc), it sounds like precise-commits solves that issue?

Exactly @cheapsteak . I also kept seeing the entire codebase getting prettier'ed and added. Instead of debugging it I just threw this in there.

PR #183, opened 2019-02-15
prettify everything

justinanastos:
The last update I just prettified `**/*`, which missed all these files.


justinanastos:
I don't know @cheapsteak :( I didn't look into it. I feel silly even putting up a PR after the last one 🤷‍♂️

Issue #182, opened 2019-02-14
Mutations crash devtools


redmerten:
Tools crash on mutations click and cache click.

 Chrome 72.0.3626.109. 
React 16.8
  "apollo-cache-inmemory": "^1.4.2",
    "apollo-client": "^2.4.12",
    "apollo-link": "^1.2.6",
    "apollo-link-rest": "^0.7.0",
    "graphql": "^14.1.1",
    "graphql-anywhere": "^4.1.27",
    "graphql-tag": "^2.10.1",

justinanastos:
Thanks for the report! The cache but is a duplicate of #107 and was resolved in [v2.1.7](https://github.com/apollographql/apollo-client-devtools/releases/tag/v2.1.7).

The mutations bug I'm working on right now and hopefully will have released by EOD.

justinanastos:
Fixed in [v2.1.8](https://github.com/apollographql/apollo-client-devtools/releases/tag/v2.1.8)

PR #184, opened 2019-02-15
Fix mutations tab

justinanastos:
The mutations tab is broken because different versions of Apollo Client return different representations of the observed mutations. Prior to v2.5, `mutation.mutation` included an AST representation of the mutation body. With v2.5, `mutation.mutationString` is included instead. This will allow for both.

Resolves issue #182

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ## vNext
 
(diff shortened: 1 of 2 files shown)


justinanastos:
You are on 🔥 with the reviews, thank you @cheapsteak !

cheapsteak:
You're on 🔥 with PRs 😄

Issue #118, opened 2018-05-04
Don't require 3rd-party cookies


nslocum:
Great work on the extension!

The extension currently requires allowing sites to set 3rd-party cookies. This setting is applied globally and is BIG deal. For me it's a deal breaker.

It looks like the Redux extension solved this issue by mocking localStorage when it's not available for whatever reason.
Here's their simple solution:
https://github.com/zalmoxisus/redux-devtools-extension/commit/493a4cc7fd37a39b28e02e9157c08e949d43a99a

MariuszHenn:
this is my solution proposal https://github.com/MariuszHenn/apollo-client-devtools/commit/9c13564c55315845e6b6ceb9835d677c4804e4b6

justinanastos:
Release in [v2.1.9](https://github.com/apollographql/apollo-client-devtools/releases/tag/v2.1.9)

Issue #142, opened 2018-10-10
Client does not appear in Chrome?


voordev:
After installing the client true Chrome extension store it does not appear as expected according to the documentation in developer tools next to console ( in that line of tabs ) 
Using latest versions as of today of all tools. 

Did I overlook something?

joeyparis:
I am experiencing the same problem. It previously worked but has since stopped and I have made no configuration changes.

Chrome: Version 69.0.3497.100 (Official Build) (64-bit)
Apollo Client DevTools: 2.1.3

hethcox:
Me too.  Chrome version 69.0.3497.100 (Official Build) (64-bit) for Mac/High Sierra.

kogai:
Same occurred for `Version 69.0.3497.100 (Official Build) (64-bit)`.

And if I have removed and installed the extension again, it works properly again, too.

hwillson:
@voordev @joeyparis @hethcox Does removing it / re-installing it help?

maimike:
no, it does not

candulabs:
Any news? It's also not working for me. Running on `Version 71.0.3578.98 (Official Build) (64-bit)` and on Apollo Client DevTools `2.1.5`. All the permissions are set properly, and I can't see any error in the extension console.

kennethlynne:
No tab shows up.

I'm using 
```
const apolloClient: ApolloClient<any> = new ApolloClient({
  cache,
  link,
  connectToDevTools: true,
});
```

```
    "apollo-cache": "^1.1.22",
    "apollo-cache-inmemory": "1.3.12",
    "apollo-client": "2.4.8",
    "apollo-link": "1.2.6",
    "apollo-link-batch-http": "1.2.6",
    "apollo-link-context": "1.0.12",
    "apollo-link-error": "1.1.5",
    "apollo-link-retry": "^2.2.8",
````

and Chrome version 71.0.3578.98 (Official Build) (64-bit)

jaybe78:
same issue for me ! anyone looking at the problem ????

karfield:
It seems resolved in latest version 2.1.6. You can clone and build it, then load the unpacked extension to get the devtool back.
@hwillson plz release the latest version to Chrome webstore asap.

smeevil:
@karfield confirmed, thanks.

anwilson514:
Has there been any update on this?  I still see 2.1.5 as the latest in the chrome store and I still cannot get the dev tools extension to work.

[...]

PR #185, opened 2019-02-19
Remove dependency on `localStorage` (and, by extension, 3rd party cookies)

justinanastos:
Chrome extensions will fail to access `window.localStorage` if 3rd party cookies are blocked. This extension uses `localStorage` in two places: in the sidebar and inside of GraphiQL.

This will replace the use of `window.localStorage` with a wrapper around [`chrome.storage.local`](https://chrome-apps-doc2.appspot.com/trunk/extensions/storage.html).

Yeah but how?

- Create an adapter between `chrome.storage` and `localStorage`.

    They have similar goals but are very different. `chrome.storage.local` uses async methods for getters and setters. `localStorage` uses sync methods. We'll keep a cached version of the local storage in memory and asynchronously sync that to chrome's storage. This removes the use of `localStorage` and the necessity of enabling 3rd party cookies. 

- Inject the storage into the application

- Share that storage throughout the application via context (so we can mock and test this later)

    Hence the React upgrade

- Consume the new storage via context

---

Resolves issue https://github.com/apollographql/apollo-client-devtools/issues/118

Hopefully contributes to issue https://github.com/apollographql/apollo-client-devtools/issues/142 🤞

### DIFF
--- a/shells/webextension/src/ChromeStorageAdapter.js
+++ b/shells/webextension/src/ChromeStorageAdapter.js
@@ -1,0 +1,96 @@
+// Expects a chrome.storage
+// [`StorageArea`](https://developer.chrome.com/apps/storage#type-StorageArea)
+// input and exports a
+// [Storage](https://developer.mozilla.org/en-US/docs/Web/API/Storage) interface
+class ChromeStorageAdapter {
+  constructor(chromeStorage, readyCallback) {
+    if (!readyCallback) {
+      throw new TypeError(
+        "ChromeStorageAdapter called without a `readyCallback`",
+      );
+    }
+    this.chromeStorage = chromeStorage;
+
+    chromeStorage.get(
+      { apolloClientDevtools: {} },
+      ({ apolloClientDevtools }) => {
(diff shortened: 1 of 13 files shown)


justinanastos:
@cheapsteak when reviewing this PR, make sure to turn off whitespace. A few of the commit wrapped stuff with new components, making the diff look large but really it was only adding a few lines.

shells/webextension/src/ChromeStorageAdapter.js
```diff
@@ -0,0 +1,83 @@
...
+    this.remoteWritePending = true;
+    this.needsRemoteWrite = false;
+
+    this.chromeStorage.set({ apolloClientDevtools: this.localCache }, () => {
+      this.remoteWritePending = false;
+
+      if (this.needsRemoteWrite) {
+        this.updateRemoteStorage();
+      }
+    });
```
cheapsteak: quite thorough!

justinanastos: I forgot to document this function. I'm going to do that before merging.

PR #193, opened 2019-03-18
Issue #187 - Scroll Fix

scfoxcode:
Fix scroll issue in chrome 72 onwards due to flex bug by applying height 100%

PR #196, opened 2019-04-01
Update inspector.less

juji:
to fix scroll bug on chrome Version 73.0.3683.86 (Official Build) (64-bit)


hwillson:
Thanks for this @juji - we've merged https://github.com/apollographql/apollo-client-devtools/pull/193 which has a similar fix (and was submitted earlier). Thanks again though!

PR #197, opened 2019-04-03
fix inspector scrolling

bennypowers:
for large caches, inspector will not scroll. this sets the max-height to 100vh


hwillson:
Thanks for this @bennypowers - we've merged https://github.com/apollographql/apollo-client-devtools/pull/193 which has a similar fix (and was submitted earlier). Thanks again though!

Issue #108, opened 2018-02-21
Queries get stuck after page refresh


tnrich:
Hey there, 

I'm finding that once a refresh occurs in my app all of the previously made queries get "stuck" in the apollo devtools queries tab. When they get stuck there no other queries get added and nothing changes as I navigate around. 

Here is a gif of this occurring from scratch. I'm on the latest version of apollo 2.

![untitled](https://user-images.githubusercontent.com/2730609/36456957-7d53e146-165c-11e8-8c1f-f63d7be02adc.gif)

As you can see in the gif, on the initial load, everything seems to be working fine, but after refreshing no other queries get added. I can confirm that this occurs for all my colleagues as well. 

I think this is very serious because it makes the devtools unusable after the initial page load.

ramsaylanier:
👍 happens to me too

seangransee:
Happens to me when refreshing the page, and when manually modifying the URL. After I manually change the URL, I have to close the Apollo dev tools and open it back up again.

wesbos:
I find my whole cache says with a page reload - I need to close dev tools and re-open it to see the correct data

wesbos:
A stop-gap solution right now is to just reload the entire frame:

![](http://wes.io/qihc/content)

deltaskelta:
using the "reload" frame option made my life a whole lot easier for now

nilshartmann:
As mentioned in #111 the problem also occures if you execute a query in the GraphiQL tab after a page reload.

The workaround from @wesbos works in that case by choosing "Reload Frame" in the `Documentation Explorer` column from GraphiQL or in the navigation tab at the left:

![image](https://user-images.githubusercontent.com/435073/41189546-dc3db722-6bcf-11e8-9e95-e5939ce6a931.png)

SeanRoberts:
Is this going to be fixed at some point?  The workaround works but it adds a lot of extra clicks/overhead when debugging

btorresgil:
Agree with other commenters, this doesn't seem to be getting enough attention given how fundamental this functionality is.  The devtools are unusable unless you reload the frame every time.  During development where refreshing the page happens every few seconds, this is extremely painful.

THPubs:
This issue is still there. Any updates on it?

ccpony:
Is anybody monitoring these issues?

justinanastos:
Thank you everyone for your report and for your patience. I'm going to be looking into this very soon.

[...]

PR #204, opened 2019-05-28
Detect devtools losing connection with apollo-client and reload

mjlyons:
This PR is a little rough, but sending as a proposal on how to deal with page refreshes (Issue #108, @justinanastos). This resolves the issue for me.

As for what we might want to clean up:
- where we store the connectionNonce on the main frame from `window.adtConnectionNonce`
- defining timeouts as consts (and maybe tweaking the values)
- maybe some of this code belongs in different locations

Happy to make changes as requested (or feel free to make them yourself), assuming the rough approach works for ya'll.

Fixes https://github.com/apollographql/apollo-client-devtools/issues/108.

### DIFF
--- a/shells/webextension/src/backend.js
+++ b/shells/webextension/src/backend.js
@@ -38,6 +38,17 @@ function handshake(e) {
       listeners = [];
     });
 
+    bridge.on("backend:set-connection-nonce", data => {
+      const {connectionNonce} = JSON.parse(data);
+      window.adtConnectionNonce = connectionNonce;
+    });
+
+    bridge.on('backend:get-connection-nonce', () => {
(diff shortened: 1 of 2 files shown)


justinanastos:
Thank you @mjlyons ! I'll try to look at this this week!

hwillson:
@mjlyons Thanks very much for working on this! Introducing a way to track and react to devtools losing a connection to Apollo Client is interesting. I think as far as addressing https://github.com/apollographql/apollo-client-devtools/issues/108 goes however, this approach adds an extra layer of complexity that isn't needed at this point. Instead, we should be able to just get away with refreshing devtools when someone refreshes their page. We'll close this for now, but will keep this approach in mind for the future. Thanks again!

PR #240, opened 2019-12-06
add navigation component to UI. introduce light/dark theme.

marthakelly:
Screen Shot 2019-12-10 at 11 56 55 AM


src/components/navigation/index.js
```diff
@@ -0,0 +1,47 @@
...
+import keymirror from 'keymirror';
+
+export const constants = keymirror({
+  'GRAPHIQL': null,
+  'QUERIES': null,
+  'MUTATIONS': null,
+  'CACHE': null
+});
+
+function setActiveItem(activeItem) {
```
marthakelly: currying (yum). lmk if this is too obtuse

static/index.js
```diff
@@ -3,6 +3,17 @@ import ReactDOM from "react-dom";
...
+const activeTheme = chrome.devtools.panels.themeName;
+
+if (activeTheme === 'light') {
+  // using require b/c import can not be synchronously dynamically loaded
+  require('./themes/light.css');
+}
+
+if (activeTheme === 'dark') {
+  require('./themes/dark.css');
+}
```
marthakelly: I tried a few more approaches here (that were syntactically elegant) but they wouldn't work. webpack doesn't like dynamic require paths, and dynamic imports are async. I could wrap a dynamic import an an async() func but i thought this might be more straightforward (?)

lmk your thoughts!

src/components/navigation/index.js
```diff
@@ -0,0 +1,47 @@
...
+import styles from './navigation.css';
+import classnames from 'classnames';
+import keymirror from 'keymirror';
+
+export const constants = keymirror({
+  'GRAPHIQL': null,
+  'QUERIES': null,
+  'MUTATIONS': null,
+  'CACHE': null
+});
```
marthakelly: is this overkill? do you prefer another lib/approach besides keymirror?

static/index.js
jbaxleyiii: Will this allow for live reloading if a users changes their preference?

src/components/navigation/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+function Navigation(props) {
+  const { GRAPHIQL, QUERIES, MUTATIONS, CACHE } = constants;
+  const defaultActiveItem = QUERIES;
+  const [active, setActive] = useState(defaultActiveItem);
+  const getClasses = setActiveItem(active);
+  const { client: { numberOfQueries, numberOfMutations } } = props;
+
+  return (
+    <div className={styles.container}>
```
jbaxleyiii: I'd love for us to explore using [Reach UI](https://reacttraining.com/reach-ui/tabs/) for these tabs so we can have some nice accessibility by default setups for core interactions

src/components/navigation/navigation.css
```diff
@@ -0,0 +1,23 @@
...
+
+.logo {
+  /* todo, get asset from torres */
+  width: 19px;
+  height: 19px;
+  padding: 13px;
+}
+
+.item {
+  color: var(--nav-item-inactive);
```
jbaxleyiii: YESSSS CSS NERD IN ME LOVES THIS

src/components/navigation/index.js
marthakelly: that is a great idea. i'll patch this and incorporate it

static/index.js
marthakelly: i think this approach won't! good call. I looked into how React handles this, and i'll try to emulate it https://github.com/facebook/react/blob/3cc564547c81b6e9a3bd6a04aa0725e668ca6559/packages/react-devtools-extensions/src/main.js#L171

marthakelly: alright @jbaxleyiii I confirmed that the entire panel is rebuilt when the theme setting is changed, and the new theme is applied

marthakelly:
@jbaxleyiii I refactored to use reach-ui (much cleaner!) and confirmed that the theme setting applies to the UI when it's set in the chrome settings. take a look at let me know your thoughts

src/devtools/background.js
```diff
@@ -1,50 +0,0 @@
...
-      reload();
-    }
-  });
-};
-
-chrome.management.getSelf(self => {
-  if (self.installType === "development") {
-    chrome.runtime.getPackageDirectoryEntry(dir => watchChanges(dir));
-  }
-});
```
marthakelly: This ended up being 1) annoying and 2) didn't really work as I expected 😬 so I'm removing it for now. I can revisit the reloading issue

PR #241, opened 2019-12-11
query list panel

marthakelly:
Screen Shot 2019-12-16 at 2 51 45 PM

* generic list panel component
* query list component
* query list item - unnamed error state
* query list item - react component state (awaiting asset from torres for react logo)
* content panels that correspond to the navigation tabs

TODO:
* search component (will complete after the right/content panel has been developed)

### DIFF
--- a/src/components/list-panel/index.js
+++ b/src/components/list-panel/index.js
@@ -15,10 +15,6 @@ function buildListItems(listItems) {
     const unnamedQuery = 'Unnamed query';
     let queryName = key;
 
-    if (key === '') {
-      queryName = unnamedQuery
-    }
-
     return (
       <li className={classnames(styles.listItem, {
         [styles.active]: index === activeListItem,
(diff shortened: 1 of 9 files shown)


src/components/list-panel/index.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+function ListPanel(props) {
+  const { listItems, listType } = props;
+  const numberOfActiveItems = Object.keys(listItems).length;
+
+  return (
+    <div className={styles.container}>
+      {/* TODO, implement search functionality. It will search over these list items and the right panel's content */}
+      <div className={styles.search}>Search {listType}</div>
```
jbaxleyiii: Probably shouldn't be a div here

marthakelly: this is a placeholder until I build out the search functionality. would you like the placeholder to be an input instead?

Issue #251, opened 2020-03-24
Apollo Devtools not Connecting using Apollo Client 3


bbrinx:
It seems that the devtools don't work with Apollo Client 3. After upgrading to Apollo 3.0.0-beta.23 I can no longer use the Apollo Devtools. The tab shows up and can be opened, but there is no content. Prior to this upgrade the devtools were hooked up and working. 

```
    "@apollo/client": "^3.0.0-beta.23",
    "@apollo/link-batch-http": "^2.0.0-beta.3",
    "@apollo/link-context": "^2.0.0-beta.3",
    "@apollo/link-error": "^2.0.0-beta.3",
    "apollo": "^2.24.0",

     Chrome v. 80.0.3987.149
```

rpCal:
Have same issue with "@apollo/client": "^3.0.0-beta.37". apollo chrome tab is not active.

onpaws:
I just ran into the same issue. 
Taking a quick look at the branches, it seems
https://github.com/apollographql/apollo-client-devtools/tree/release-3.0
might be 3.0 compatible.

onpaws:
I was able to get `release-3.0` built and installed in Chrome, it looks like this:
![image](https://user-images.githubusercontent.com/182515/78670937-d59b0280-78de-11ea-9e9c-5c139e240775.png)
The queries appear to be hard coded (not from my app). Seems it may still be too early to use.

@justinanastos, @cheapsteak not sure if you're still maintaining? 
Would be appreciated if you could provide an update. Thanks for your work on this. 🍻

jcreighton:
@bbrinx Thanks for opening this issue! Changes in the AC3 beta are incompatible with the current extension but we are working on a fix! 

@onpaws I'll be leading on this project 😀Looking forward to getting you AC3-compatible devtools!

onpaws:
Glad to hear it @jcreighton! Good luck. 
If memory serves I heard the Firefox extension was blocked by Mozilla for using `eval`, so if that's true, I especially wish you luck in that endeavour! :)

prins-p:
@jcreighton , that's great, thank you!! Could you give us some guidance when you expect to be able roll out an update of the devtools with AC3 support?

PR #263, opened 2020-05-25
Apollo Client 3 (@apollo/client) updates

hwillson:
This PR updates devtools to be compatible with both Apollo Client 2 and 3. The commit messages go into more details, but there is one important thing to call out. Since we're working on a new version of devtools that is essentially a rewrite of the existing codebase, I've done the minimum necessary to get the existing devtools working with AC3. In most cases this is okay, but it does mean there are slight differences/limitations in the "Cache" panel. From https://github.com/apollographql/apollo-client-devtools/commit/5b256d6a8b45390597b24c9a7ff2c9d23fac2602: 

> We're working on a complete re-write of the AC devtools, so this
> commit should be considered temporary.
> 
> Apollo Client's internal cache structure has changed between
> AC2 and AC3. This means the `Inspector` code bits of devtools
> (that drive the Cache panel) are no longer compatible with the
> newer cache structure. The current `Inspector` relies on AC2's
> internal cache structure to render a hierarchical view of the
> cache, with links between different entities. Re-creating this
> for AC3 will take time away from our newer devtools rewrite
> work, so this commit focuses on making the minimal changes
> needed to get the Cache panel working with AC3.
> 
> With this commit, AC3 cache contents are displayed in the
> Cache panel, but contents are not properly linked (e.g. cache
> references are not followed and linked together). Nested
> entities are displayed as simple JSON dumps.

In other words, this is how the "Cache" panel looks/behaves when used with AC2:

Image 2020-05-25 at 1 58 25 PM png

Whereas this is how it looks when used with AC3:

Image 2020-05-25 at 1 59 16 PM png

The AC3 version dumps a full representation of the cache contents, but it does not automatically link in referenced entities. The devtools rewrite we're working on will take care of all of this (and more), but hopefully this will serve as a stop gap until the new devtools are ready.

Fixes https://github.com/apollographql/apollo-client-devtools/issues/251.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "apollo-client-devtools",
-  "version": "2.2.4",
+  "version": "2.2.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/backend/broadcastQueries.js
+++ b/src/backend/broadcastQueries.js
@@ -1,3 +1,21 @@
+// Apollo Client 2 stores query information in an object, whereas AC3 uses
+// a Map. Devtools are expecting to work with an object, so this function
+// will convert an AC3 query info Map to an object, while also filtering out
+// the query details we don't need.
+function filterQueryInfo(queryInfoMap) {
+  const filteredQueryInfo = {};
+  queryInfoMap.forEach((value, key) => {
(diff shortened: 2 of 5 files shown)


Leo4815162342:
@benjamn shall we merge this temp solution?

PR #274, opened 2020-06-03
Restructure codebase + new build process + upgraded packages

jcreighton:
### What's in this PR

📁 New folder structure

All extension-related files live in `src/extension` and are further grouped to mirror the architecture of the extension itself. In the previous architecture, it was difficult to keep track of what files were where, how those files related to each other and where those files "lived" (Is this file running as a content script in the inspected tab? Or no, wait, is it part of the devtools panel?) I'm hoping this new structure provides some clarity as files are now grouped by whether they "live" in the `tab`, `background` or `devtools`. Files were also renamed for clarity.

The application now lives under `src/application`.

As part of the restructuring, the `shells` directory has been removed. 

🛠 New build process

The new process outputs to the `build` folder where you can view the built files in the same representation as what we upload to the Firefox and Chrome extension stores. In development, this includes source maps. In production, source maps are not included and the files are minified. The distributable files are now created from the `build` folder (but continue to be zipped and output to `dist`).

🏗 Upgraded packages

Packages are upgraded and unused packages are removed. Huzzah!

### What's coming next

🎢 Improved application + extension architecture

I purposefully limited the scope of this PR to re-arranging the structure and build. I'll be holding off on improvements to the actual application or extension until the items below are in place.

🚀 Improved development process

I'll be focusing on creating the most pain-free development process I can. Extension development has some limitations but for Firefox, we have the awesome `web-ext` tool to take advantage of. I'll also add some test applications we can boot up and inspect.

🚦 TypeScript + tests

TypeScripts and tests are our friends.

🦄 ESLint

Previously, we used Prettier. I'm unsure if I'll add it back but ESLint is 100% needed (and in particular if we use hooks in the application side of things).


README.md
```diff
@@ -51,47 +51,54 @@ If you're seeing the "Apollo" tab but still having issues, skip ahead to the **D
...
+```bash
+npm run firefox
+```
+Now, while on any page, open the inspector. If you're inspecting a page that is using Apollo Client, there will be a global `window.__APOLLO_CLIENT__` object on that page. If that object exists, you will see an "Apollo" tab in the inspector menu. This tab will contain the Apollo Client devtools.
 
-### Reloading the Chrome extension
+### Folder structure
 
-Unfortunately, there is no way to hot-reload a Chrome extension in the inspector while developing it.
+There are two main pieces of the Apollo Devtools: the extension itself and a React application. The extension is the code that communicates with the browser. It's what allows us to search an inspected window for an instance of Apollo Client and to create the Apollo tab in the browser's devtools panel. The React application is what powers the experience what the user clicks on the Apollo tab.
```
hwillson, line 89: The wording seems a bit off in this sentence: "The React application is what powers the experience what the user clicks on the Apollo tab."

jcreighton: Nice catch! I'm going to fix the grammatical error there. But otherwise, leave it as-is because this README is 100% going to be revised in the near future.

PR #276, opened 2020-06-29
🚀 Improved development process

jcreighton:
As mentioned in https://github.com/apollographql/apollo-client-devtools/pull/274, this pull request encompasses the work to create a better, more pain-free development process, including:

- A sample application to develop (and potentially test) against
- Commands to spin up Firefox *and* Chrome with the extension temporarily installed
- The same commands also watch and rebuild the files + reload the extension 💃


development/client/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "colors",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@apollo/client": "^3.0.0-rc.4",
```
hwillson, line 6: Maybe just replace `^3.0.0-rc.4` with `latest`, as we've already jumped up to `3.0.0-rc.10` in a few days, and there are more coming! 😂

development/client/src/App.js
```diff
@@ -0,0 +1,51 @@
...
+import ColorSchemeGenerator from './ColorSchemeGenerator';
+import Favorites from './Favorites';
+import ColorLookup from './ColorLookup';
+import { GET_SAVED_COLORS } from './queries';
+import './App.css';
+
+function App() {
+  const inputRef = useRef(null);
+  useQuery(GET_SAVED_COLORS);
+  const client = useApolloClient();
```
hwillson, line 18: It looks like this line (and the `useApolloClient` import) can be removed (`client` isn't being used).

development/client/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+              });
+            }
+          },
+        }
+      }
+    },
+  }),
+  link: new HttpLink({
+    uri: 'http://localhost:4000',
+  }),
```
hwillson, line 30: With AC3 this can now just be

```suggestion
  uri: 'http://localhost:4000',
```

 (the link is automagically created behind the scenes)

package.json
```diff
@@ -25,8 +32,11 @@
     "css-loader": "^3.5.3",
     "less": "^2.7.1",
     "less-loader": "^6.1.0",
+    "npm-run-all": "^4.1.5",
+    "rimraf": "^3.0.2",
     "style-loader": "^1.2.1",
     "terser-webpack-plugin": "^3.0.2",
+    "ttab": "^0.6.1",
```
hwillson, line 39: Is `ttab` being used by the app, or is this just a convenience dependency? I didn't see it being referenced anywhere else.

jcreighton: Accidentally left it in when I was trying something out!

development/client/package.json
jcreighton: 🚀 Good idea! Done!

PR #279, opened 2020-07-23
Reconfigure data flow

jcreighton:
**Is Relay...uh... _that_ Relay?**

It's not. I know the name I've chosen here is... an interesting choice. But also it's what that abstraction does, so 🤷‍♀️ . Relay in this case is an abstraction that allows you to add connections and send messages to those connections which might also send those messages to _other_ connected Relays. 

**But why?**

Previously, we relied on the Bridge (a simple interface that extended from EventEmitter) to send messages back and forth. But that abstraction had a few issues, the biggest of which is you weren't sure where the messages were coming from or where those messages were ending up. A bridge might talk to another bridge that talked to another bridge. Relays are meant to clarify those connections. When you see a Relay is sending a message, it will explicitly state the path for that message, such as `tab:background:devtools` or `background:tab`. I hope this removes some confusion and mental load when working with this project.

**What else?**

Using Relay, I've reconfigured how the devtools are initialized + how data flows from the client (aka userland) to the devtools. _Also_, the devtools application now uses Apollo Client to store the data from the client. Previously, that data flowed into the application through React's Context... but Context isn't great for storing more than a single value. Objects (and nested objects) stored in Context often cause unintended re-renders. Admittedly, this is a strange use case for Apollo Client but so far, so good. 🤞


src/application/index.js
```diff
@@ -1,36 +1,59 @@
...
-        <StorageContextProvider storage={shell.storage}>
-          <Panel
-            isChrome={isChrome}
-            bridge={bridge}
-            theme={isDark ? "dark" : "light"}
-          />
-        </StorageContextProvider>
-      </BridgeProvider>
-    );
+const queriesVar = cache.makeVar(null);
```
jcreighton, line 24: Using reactive vars to store the data. At first, I used `writeQuery` but when querying for the `cache` object, I received a lot of errors. Guessing that there are some issues with returning a cache in a cache which makes sense. I still had a bit of trouble with the cache object here but JSON.stringifying it worked like a charm. 

I'm cool with how this is being handled here but if you see a potential issue or a better option, please let me know. 🙏

src/extension/tab/hook.js
```diff
@@ -70,28 +13,63 @@ export function installHook(window, devToolsVersion) {
...
 
-  interval = setInterval(findClient, 1000);
+  // Attempt to find the client on a 1-second interval for 10 seconds max
+  findClient();
+
+  window.addEventListener('message', ({ data }) => {
+    if (data?.message === 'devtools-initialized') {
+      if (hook.ApolloClient) {
+        window.postMessage({
+          message: 'create-devtools-panel',
```
benjamn, line 50: Let's use the constant `CREATE_DEVTOOLS_PANEL` here, rather than its value.

src/extension/tab/hook.js
```diff
@@ -70,28 +13,63 @@ export function installHook(window, devToolsVersion) {
...
+          }),
+        });
+      }
+    }
+
+    if (data?.message === 'request-update') {
+      window.postMessage({
+        message: 'update',
+        to: 'tab:background:devtools',
+        payload: JSON.stringify({
```
benjamn, line 65: It shouldn't be necessary to `JSON.stringify` the payload, since `postMessage` can serialize basic objects. If there's something that `postMessage` can't handle, I wouldn't trust `JSON.stringify` to handle it any better.

Do you recall why this stringification helped?

src/extension/tab/tabRelay.js
```diff
@@ -0,0 +1,45 @@
...
+
+async function requestId() {
+  const promise = new Promise((resolve) => {
+    chrome.runtime.sendMessage({ message: 'request' }, async function(response) {
+      resolve(response);
+    });
+  });
+
+  return await promise;
+}
```
benjamn, line 11:
```suggestion
function requestId() {
  return new Promise((resolve) => {
    chrome.runtime.sendMessage({ message: 'request' }, response => resolve(response));
  });
}
```
No need for `async`/`await` when you can just return a `Promise`.

src/extension/devtools/panel.js
```diff
@@ -1,57 +1,6 @@
...
-  `;
-  chrome.devtools.inspectedWindow.eval(src, function(res, err) {
-    if (err) console.log(err);
-    cb(res);
-  });
-}
+window.devtools = {
+  initialize: initDevTools,
+  writeData,
+};
```
benjamn, line 6: Since it seems possible to `import` these functions from `../../application`, I'd like to avoid hanging them off the global object, and stick to `import`ing them wherever they're needed.

src/extension/tab/backend.js
```diff
@@ -0,0 +1,10 @@
+import relay from './tabRelay';
+import { initializeHook } from "./hook";
+import { version as devToolsVersion } from "../manifest.json";
+
+if (document instanceof HTMLDocument) {
```
benjamn, line 5:
```suggestion
if (typeof document === "object" && document instanceof HTMLDocument) {
```
Is there any chance `document` might not be defined? If so, you should probably use `typeof` to check beforehand.

src/extension/tab/backend.js
```diff
@@ -0,0 +1,10 @@
...
+  const script = document.createElement("script");
+  script.textContent = `;(${initializeHook.toString()})(window, "${devToolsVersion}")`;
```
benjamn, line 7: Any reason we can't `import { initializeHook } from './hook'` and call it like the function that it is? As long as it gets the right `window`, I think it should work?

src/extension/tab/hook.js
```diff
@@ -70,28 +13,63 @@ export function installHook(window, devToolsVersion) {
...
-        },
-      );
-      clearInterval(interval);
+    let interval;
+    let count = 0;
+  
+    function initializeDevtoolsHook() {
+      if (count++ > 10) clearInterval(interval);
+      if (!!window.__APOLLO_CLIENT__) {
+        hook.ApolloClient = window.__APOLLO_CLIENT__;
```
benjamn, line 30: Not urgent, but we should think about how we might be able to handle more than one client instance.

src/extension/tab/hook.js
```diff
@@ -70,28 +13,63 @@ export function installHook(window, devToolsVersion) {
...
+        hook.ApolloClient.__actionHookForDevTools(handleActionHookForDevtools);
+        hook.getQueries = () => hook.ApolloClient.queryManager.queries();
+        hook.getMutations = () => hook.ApolloClient.queryManager.mutationStore.getStore();
+        hook.getCache = () => hook.ApolloClient.cache.extract(true);
```
benjamn, line 34: At some point we may want to distinguish between optimistic data (`cache.extract(true)`) and realistic data (`cache.extract(false)`), so that people can see how their optimistic mutations are working.

src/Relay.ts
```diff
@@ -0,0 +1,103 @@
...
+  connections = new Map();
+
+  constructor(name:string) {
+    super();
+    this.name = name;
+    this.broadcast = this.broadcast.bind(this);
+    this.send = this.send.bind(this);
+    this.addConnection = this.addConnection.bind(this);
+    this.removeConnection = this.removeConnection.bind(this);
+    this.createEvent = this.createEvent.bind(this);
```
benjamn, line 20: Instead of binding methods in the constructor, you can use arrow function syntax when defining the method below:
```ts
broadcast: (message: Message, sender: ???) => {
  ...
}
```
That said, if you don't plan on passing the method around as a function, then you don't need to bind it, which means it can live on `Relay.prototype`, rather than getting redefined on each `Relay` object.

src/Relay.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  dispatch(message) {
+    this.dispatchEvent(message);
+  }
+
+  createEvent(message:string, detail:Message = {}) {
+    return new CustomEvent(message, { detail });
+  }
+
+  broadcast(message:Message, sender) {
```
benjamn, line 48: What's the type of `sender` here? I think you'll be glad you went to the trouble of specifying types for everything you can in this module, since it's so central to everything you're doing.

src/extension/devtools/devtools.js
```diff
@@ -1,72 +1,67 @@
...
+} from '../constants';
 
-let panelLoaded = false;
-let panelCreated = false;
-let panelShown = false;
-// stop after 10 seconds
-let checkCount = 0;
-let loadCheckInterval;
+const inspectedTabId = chrome.devtools.inspectedWindow.tabId;
+const devtools = new Relay('devtools');
```
benjamn, line 9: It seems a bit confusing to have both `window.devtools` and a local variable called `devtools`. That said, I think/hope we can get rid of `window.devtools` (see other comments), so this should sort itself out?

src/application/index.js
benjamn: This works, but it definitely feels a little bit "because we can"-ish. However, I don't immediately have a simpler suggestion, so let's go with this for now!

development/client/tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "react",
+    "allowSyntheticDefaultImports": true
```
hwillson, line 20: Since `esModuleInterop` is already enabled, you shouldn't need `allowSyntheticDefaultImports`.

jcreighton: Oh, it got taken out in the `testing` branch that got merged so I'll make sure it gets removed here too.

src/Relay.ts
jcreighton: I can convert these to arrow functions. I wasn't sure what style we preferred but it doesn't matter much to me.

src/Relay.ts
```diff
@@ -0,0 +1,103 @@
+interface Message {
+  to?: string
```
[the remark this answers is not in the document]
jcreighton: Yes, if there's one recipient. But we might send messages to two+ recipients, in which case I prefer `['one:two', 'one:two:three']` to `[['one', 'two'], ['one', 'two', 'three']]`. Still waiting to see if that's the case though as I build the rest of this out. If it's not, I'm cool with switching to this!

src/extension/background/background.ts
```diff
@@ -0,0 +1,28 @@
+import Relay from '../../Relay';
+import { REQUEST_TAB_ID } from '../constants';
+
+// This sends the tab id to the inspected tab.
+chrome.runtime.onMessage.addListener(({ message }, sender, sendResponse) => {
+  console.log('for never leaving well enough alone');
```
jcreighton, line 6: Please ignore these Taylor Swift lyrics. Will remove shortly.

jcreighton:
@benjamn I flipped `strictNullChecks` to true and fixed the resulting errors. But... I might flip it back as I work on the React application. The plan for the TypeScript conversion is to start loose and then tighten up as we go along, so depending on how many errors that produces in that code... well 😅 we'll see what happens.

benjamn:
@jcreighton I'm still seeing errors like
```
src/extension/devtools/devtools.ts:25:26 - error TS2345: Argument of type '{ to: string; }' is not assignable to parameter of type 'Message'.
  Property 'message' is missing in type '{ to: string; }' but required in type 'Message'.

25   devtools.send(message, {
                            ~
26     to: `background:tab-${inspectedTabId}:client`
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
27   });
   ~~~
```
when I run `npx tsc`. If you're just using VSCode, sometimes it doesn't show errors for files you don't have open, but `tsc` will catch everything.

In this case, I wonder if you could include `message` as a property in the `Message` argument, instead of passing it as a separate parameter? Then `send` could pass that `Message` object through to `this.broadcast`, ideally?

Instead of changing `strictNullChecks` to `false` again, you should be able to have multiple `tsconfig.json` files for different subdirectories, and they can even [extend](https://www.typescriptlang.org/tsconfig#extends) each other. That said, the sooner you can resolve these strictness checks, the better off you'll be in the long run. Of all the things that TypeScript complains about, this family of errors is one of the most useful, compared to writing plain JS, I believe.

tsconfig.json
```diff
@@ -2,11 +2,12 @@
   "compilerOptions": {
```
benjamn, line 2: I would recommend adding `"outDir": "./dist"` to `compilerOptions`, like we do for Apollo Client [here](https://github.com/apollographql/apollo-client/blob/610293d091f006e8044ae9853822bb96cb261756/tsconfig.json#L18), so that the compiled files don't end up mixed together with the `src/*` files. Then you can add `dist/` to `.gitignore` to avoid seeing all the compilation artifacts every time you run `git status`.

src/__tests__/Relay.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      to: undefined,
+    });
+
+    relay.removeConnection('Bedford Ave');
+
+    relay.send('The L train is still delayed.', {
+      to: 'Bedford Ave',
+    });
+
+    expect(callback).not.toBeCalledTimes(2);
```
benjamn, line 54: Speaking from painful experience, tests that expect things to happen synchronously can become a real liability down the road, especially when it's something like sending/receiving a message, which you might want to be asynchronous at some point in the future.

A different approach is to use the `resolve` function from `new Promise(resolve => ...)` as the callback, and return a `Promise` from the test, so the test will pass if/when the `resolve` function is called, or time out after 5000ms if it isn't called. It may also be tempting to use `async`/`await`, but that doesn't work so well when there are callback functions involved.

src/Relay.ts
```diff
@@ -0,0 +1,78 @@
+interface Message {
+  to?: string | undefined
+  message: string
+  payload?: any | undefined
```
benjamn, line 4: There's a slight difference between an optional (`?:`) field and one that can be `undefined`. There are definitely use cases for each style, but you don't usually need to use both together. In this case, I would say that the optional fields don't need `| undefined`, since that's implied by the optionality. Also, `any` is already a sinkhole for type-safety, so `any | undefined` might as well be `any`.

Here's my recommendation:
```ts
interface Message<TPayload = any> {
  to?: string;
  message: string;
  payload?: TPayload;
}
```

src/extension/tab/hook.js
benjamn: I still think it would be good to find a way to avoid `JSON.stringify` if possible!

tsconfig.json
jcreighton: I pushed up a commit that includes `ts-loader` so running `npm run build` will now catch those errors. From what I understand, we don't need to run `tsc` to compile as Webpack is doing that for us (now) so I don't believe we need an `outDir`. TBH I've wanted to chuck this laptop out the window *multiple* times with configuring TS + Babel + Webpack. I *think* it's working now but it's some kind of dark magic.

src/extension/tab/hook.js
jcreighton: I missed removing these earlier. But it look like there's an issue with cloning the following:
Screen Shot 2020-08-11 at 10 02 59 PM

Is that expected? I might be doing something wrong here.

jcreighton:
@benjamn Oh, that's a great idea!

PR #289, opened 2020-09-19
Initial emotion and @apollo/space-kit config

hwillson:
This PR wires in `emotion` and `@apollo/space-kit`, and temporarily demonstrates that they're working. Note that it also completely removes `less`, and all `.less` style files. If we need those styles back for any reason while we're working on individual components, we can always pull them back from GH. I'm assuming we won't need them though due to our plans to leverage space-kit as much as possible.


src/application/index.tsx
```diff
@@ -1,6 +1,11 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core";
```
hwillson, line 2: Note that I'm leveraging `emotion`'s recommended pragma approach here, instead of chasing dragons through their babel config approach. The babel config option is super flakey when working with typescript and `ts-loader`, so I've dropped it. We can re-visit it in the future as needed, but for now adding `/** @jsx jsx */` and importing `jsx` are required in each file that uses `emotion`.

jcreighton: What a bummer. But 🤷‍♀️, we can live with it.

jcreighton (approved):
LGTM

Thanks, @hwillson!

PR #291, opened 2020-09-28
Initial Circle CI setup/integration

hwillson:
Wire in Circle for linting/testing. Note that the lint job has been disabled for now, due to it returning a lot of errors on the old code we're planning on removing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,32 @@
+version: 2.1
+orbs:
+  node: circleci/node@4.0.1
+
+jobs:
+  lint:
+    executor:
+      name: node/default
+      tag: 'current'
+    steps:
+      - checkout
+      - node/install-packages
+      - run:
+          command: |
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+  lint:
+    executor:
+      name: node/default
+      tag: 'current'
+    steps:
+      - checkout
+      - node/install-packages
+      - run:
+          command: |
+            npm run build:prod
```
jcreighton, line 15: Should this be `build`? `build:prod` minifies the code.

hwillson: Happy to adjust! I used `build:prod` to ensure we were testing code that's a close to our prod deployed code as possible, but I agree it isn't really necessary in this case (and skipping minifying will speed things up a bit). Thanks!

jcreighton (approved):

PR #295, opened 2020-10-06
Create Layout components with Navigation

jcreighton:
Our design share one common component, Navigation, but two different layout types, one with a sidebar and one utilizing the full width of the screen. This PR adds the Navigation, SidebarLayout and FullWidthLayout components to the application. You'll also find skeleton components for the Queries and Mutations screens, as well as a slight re-arrangement in the Explorer screen.


src/application/Layouts/FullWidthLayout.tsx
```diff
@@ -0,0 +1,65 @@
...
+  navigationProps: NavigationProps
+  children: any;
+};
+
+interface FullWidthLayoutComposition {
+  Header: React.FC;
+  Main: React.FC;
+}
+
+const layout = css`
```
hwillson, line 17: Coding preference question: do you think we should name our serialized emotion styles with a style suffix of some kind, to differentiate them a bit from other vars in our source files? E.g. instead of `layout` we use something like `layoutStyles` or `layoutCss`, just to differentiate them a bit at a quick glance from other non-css based vars.

hwillson (approved):
This looks awesome @jcreighton - and tests, woohoo! I just have one quick question, but feel free to disagree and ignore. LGTM - thanks!

src/application/Layouts/FullWidthLayout.tsx
jcreighton: Oh, yes! That's a great idea. I'll update soon-ish.

PR #299, opened 2020-10-26
Inspect watched queries in the Queries panel

jcreighton:
What this PR covers:

- User can view watched queries
- User can select a watched query to view its query string, variables, and cached data
- User can copy the query string, variables, or cached data
- Testing utilities moved into `utilities/testing` 

Screen Shot 2020-10-20 at 1 13 23 PM

What it doesn't cover:

- User cannot select the "Run in GraphiQL" button (this will be handled in a separate PR)
- Styling between the query string, variables/cached data differs (to be handled separately)
- UI work to handle scrolling when queries list is longer than UI (same for the cached data) and other UI tweaks


src/application/Queries/Queries.tsx
```diff
@@ -1,22 +1,135 @@
...
+      queryString
+      variables
+      cachedData
+    }
+  }
+`;
 
 export const Queries = ({ navigationProps }) => {
+  const [selected, setSelected] = useState<number>(0);
+  const theme = useTheme<any>();
```
hwillson, line 87: To get rid of this `any`, you could narrow the types used in the `theme.ts` file a bit, then replace `any` with something that will point to one of the theme objects. In other words, you could update `src/application/theme.ts` to use `ShadedColor` as the second `Record` value type:

```ts
import { lighten } from "polished";
import { colors, ShadedColor } from "@apollo/space-kit/colors";

export enum ColorTheme {
  Light = 'light',
  Dark = 'dark'
}

const shared = {
  sidebarSelected: colors.indigo.base,
};

export const themes: Record<ColorTheme, Record<string, ShadedColor>> = {
  [ColorTheme.Light]: {
    primary: colors.indigo.darkest,
    sidebarHover: lighten(0.2, colors.indigo.darkest) as ShadedColor,
    ...shared,
  },
  [ColorTheme.Dark]: {
    primary: colors.black.base,
    sidebarHover: lighten(0.2, colors.black.base) as ShadedColor,
    ...shared,
  },
};
```

then in `Queries.tsx` reference the theme object with something like:

```ts
import { themes } from "../theme";
...
type Theme = (typeof themes)[keyof typeof themes];
```

then use this new `Theme` type like:

```ts
export const Queries = ({ navigationProps }) => {
  ...
  const theme = useTheme<Theme>();
```

src/extension/tab/hook.ts
```diff
@@ -24,7 +24,7 @@ declare global {
 type Hook = {
   ApolloClient: ApolloClient<TCache> | undefined;
   version: string;
-  getQueries: () => void;
+  getQueries: () => Map<string, any> | void;
```
hwillson, line 27: You should be able to replace this `any` with `QueryInfo` (see comment below).

src/extension/tab/hook.ts
```diff
@@ -54,7 +54,39 @@ function initializeHook() {
...
   });
 
+  type QueryInfo = {
+    document: any;
+    source: any,
+    variables: any;
+    lastWrittenResult: any;
+    lastWrittenVars: any;
+  }
+
```
hwillson, line 64: Maybe consider moving this out of the `initializeHook` function, putting it above the `Hook` type declaration above. That way `Hook` can use this as part of its `getQueries` definition (see comment above).

Also, to help get rid of the `any`'s here, maybe something like:

```ts
import { DocumentNode, Source } from "graphql";
...
type QueryInfo = {
  document: DocumentNode;
  source?: Source,
  variables?: Record<string, any>;
  lastWrittenResult?: Record<string, any>;
  lastWrittenVars?: Record<string, any>;
}
```

src/extension/tab/hook.ts
```diff
@@ -54,7 +54,39 @@ function initializeHook() {
...
+    const queryMap = hook.getQueries();
+    let queries: QueryInfo[] = [];
+
+    if (queryMap) {
+      queries = [...queryMap.values()].map(({ 
+        document, 
+        variables,
+        lastWrittenResult,
+        lastWrittenVars,
+        diff,
```
hwillson, line 76: I think you can drop the `diff` here and below.

hwillson (approved):
This looks awesome @jcreighton! (and sorry for the review delay) 🎉 I've added a few small TS suggestions, but 👍 overall. Thanks!

PR #300, opened 2020-10-26
Release 2.3.2

jcreighton:
Includes a fix for discovering local-only schemas (thank you @micmro!)


jcreighton:
@hwillson Am I correct that we can ignore the failing tests here?

hwillson:
@jcreighton looks like it - thanks!

PR #286, opened 2020-09-09
Fix the local state functionality in the dev tools

micmro:
Since #263 the client-only schemas are not passed from the backend to the dev-tools anymore, which caused the dev-tools to not show any local-only schema information.

This PR is adding this back in, but using an immutable array as the previous version would append another `apolloClientSchema` copy to `schemas` each time `next` was called.

Test setup:
```typescript
import { ApolloClient, gql, InMemoryCache } from "@apollo/client";

const typeDefs = gql`
  extend type Query {
    testClientOnlyFiled: [String]
  }
`;

const cache = new InMemoryCache({
  typePolicies: {
    Query: {
      fields: {
        testClientOnlyFiled: {
          read() {
            return "TEST";
          },
        },
      },
    },
  },
});

export const apolloClient = new ApolloClient({
  uri: "https://graphqlzero.almansi.me/api",
  cache,
  typeDefs,
});
```

Before this change (expected `testClientOnlyFiled` to be in _Query_ and _Explorer_ panel):
Screenshot 2020-09-09 at 10 42 45 PM

After this change (`testClientOnlyFiled` is in _Query_ and _Explorer_ panel):
Screenshot 2020-09-09 at 10 40 42 PM

PR #307, opened 2020-10-30
Fix error caused by undefined schemas

jcreighton:
Issued with #286 - we didn't check that `schemas` existed.

I am grateful that we'll be able to avoid bugs like this in the future codebase. 🙏


hwillson (approved):
👍 thanks @jcreighton!

PR #310, opened 2020-11-05
Inspect mutations in the Mutations panel

jcreighton:
Similar to #299, we can now inspect mutations in the Mutations panel! 

Users can:

- View the list of mutations
- Select a mutation to view its mutation string and variables
- Copy the mutation string or variables
- Use the Run in GraphiQL button to run the mutation (or query if on Queries) in Explorer

Some extra work was done to ensure overflowing sidebar, operation string, and operation data are scrollable.

Screen Shot 2020-11-05 at 4 11 06 PM


hwillson (approved):
This looks AMAZING @jcreighton - LGTM - thanks!

PR #311, opened 2020-11-20
Handle data requests and tab reloads

jcreighton:
This PR sets our data requests at a 500ms interval. We'll explore further performance optimizations in the future. We also now handle a tab reload _without refreshing the devtools application_. 🎉


hwillson (approved):
This looks amazing @jcreighton - thanks very much!

PR #315, opened 2020-12-09
Enable Dark Mode 🌑

jcreighton:
This PR enables Dark Mode based on [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

Some styles changed in this PR but most styles changes will be in a separate PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9529,6 +9529,12 @@
         }
       }
     },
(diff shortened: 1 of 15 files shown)


hwillson (approved):
Looks awesome @jcreighton - thanks!

PR #316, opened 2020-12-10
Hide introspection query

jcreighton:
The IntrospectionQuery is part of the setup for GraphiQL but we don't want it to show on the Queries panel. We now filter it out.


hwillson (approved):
👍 @jcreighton - thanks!

PR #318, opened 2020-12-10
Handle mutations for Apollo Client 3.3

jcreighton:
How mutations are retrieved changed in Apollo Client 3.3. This change allows us to handle both Apollo Client 3.0 - 3.2 and 3.3+ in the devtools.


hwillson (approved):
🎉 thanks @jcreighton!

PR #320, opened 2020-12-11
Delete unused components from v2

jcreighton:
Let us take a moment of silence to thank these components for their work. _a somber head bow_ Farewell, dear friends.

On a more serious note, we no longer use these components in v3 so we're deleting them.


hwillson (approved):
Woohoo! 👋 🥳

Issue #294, opened 2020-09-30
Apollo Dev Tools don't refresh data


wesbos:
I'm not able to see whats in my cache via apollo dev tools. I can refresh the frame (which only exists on Chrome, not firefox), but about half the time it can't find the cache.

I can see the cache if  I run `__APOLLO_CLIENT__.cache.data.data` in the console, so the data is there.

It's almost like the dev tools aren't aware of changes to cache most of the time. For some reason, sometimes changing the page will cause it to update. 

[Here is a video detailing the problem](https://wes.io/6quP7dyA), hopefully it's helpful. Any tips on a workaround for now?

wesbos:
Just seeing this PR for a new version of dev tools, I bet this will fix it. Any way to beta test it?

https://github.com/apollographql/apollo-client-devtools/pull/292

jcreighton:
Hi @wesbos! So, as you've seen, we're working on an overhaul of the existing devtools. We're not ready for beta testing yet but we'd love your feedback when we are! 

Re: "It's almost like the dev tools aren't aware of changes to cache most of the time" - we're working to make 3.0 more reliable. We won't be working on fixing issues in 2.x so we can focus on getting 3.0 out the door.

wesbos:
Thanks. Would you merge a hot fix if I can at least make 2.x work for now? It’s unusable right now

jcreighton:
@wesbos Would be happy to!

jcreighton:
Good news! I was able to find the issue here and a fix is coming with 2.3.2! Please check out the associated PR and confirm the fix.

wesbos:
amazing! Thank you so much :) Will try it out shortly once i'm back on that project

jcreighton:
I went ahead with publishing 2.3.2 in Chrome and Firefox 👍

jcreighton:
Closing this as the latest version should fix this error. Feel free to re-open if that's not the case!

wesbos:
@jcreighton finally getting around to testing this and it doesn't seem to be any different in Firefox or Chrome (2.3.3 on both)

The cache does not show any updates unless I change the page, or close devs tools.  Here is a video: https://wes.io/QwubK1v7

wesbos:
How do I re-open? Make another issue?

jcreighton:
[...]

PR #321, opened 2020-12-12
Fetch data every 500ms unless data is already enqueued

jcreighton:
Investigating #294 revealed that the cache doesn't update on longer-running requests. It works as expected for fast requests.   I've modified the hook to request data every 500ms (as long as an update is not enqueued). That seems to be working, so hopefully this fixes the issue.


hwillson (approved):
Thanks @jcreighton!

PR #322, opened 2020-12-13
Add additional UI tests for the cache panel

hwillson:
As well as a few adjustments for an empty cache state.


jcreighton (approved):
LGTM!

PR #323, opened 2020-12-14
Use EventTarget shim for Firefox compatibility

jcreighton:
Turns out, the native EventTarget interface doesn't work in content scripts in Firefox. We'll use a shim to provide the functionality we need.


hwillson (approved):
Great catch @jcreighton - thanks!

PR #327, opened 2020-12-16
Dark Mode with CSS Variables

jcreighton:
This PR removes `emotion-theming` and introduces CSS variables to handle Dark and Light Modes. It also corrects some styles on Dark Mode to make it more usable. We'll have a more in-depth review of Dark Mode with design in the near future so please consider these design changes temporary.

DarkMode


hwillson (approved):
This is awesome @jcreighton - LGTM! 👍

Issue #328, opened 2020-12-16
Extension creates infinite loop on some website


jcreighton:
**How to reproduce the issue:**
Visit Glitch.com with the latest version of the devtools. The loader should spin and eventually crash the site.

**Desktop (please complete the following information):**
 - Extension version: 3.0

jcreighton:
Related to this issue on the event-target-shim repo: https://github.com/mysticatea/event-target-shim/issues/26

We'll need to replace the shim with another package as this doesn't look likely to be resolved soon.

PR #330, opened 2020-12-18
Replace EventTarget with simplified implementation 

jcreighton:
As noted in #323, we replaced EventTarget with a shim due to an issue with EventTarget not dispatching events in the content script. The issue affected Firefox but not Chrome. That caused an unintentional issue noted in #328 that resulted in infinite looping on websites using Sentry. This PR replaced the shim with a simplified EventTarget implementation.


hwillson (approved):
Great fix for a bizarre issue - thanks @jcreighton!

PR #332, opened 2020-12-18
Empty states for Queries & Mutations panels

jcreighton:
This PR changes the Queries & Mutation panels to show an empty state when no queries or mutations are present.


hwillson (approved):
Awesome @jcreighton - thanks!

PR #410, opened 2021-02-16
Perform selected operation.

slorenzo:
This PR aims to handle the execution when there are 2 or more operations on the Graphiql editor. 

Discussion: https://github.com/apollographql/apollo-client-devtools/discussions/405


hwillson (approved):
This is really great @slorenzo - thanks very much!

PR #446, opened 2021-03-12
Cache EntityView bugfix: null or boolean values are not rendered

alexTayanovsky:
When viewing an entity in the "Cache" section, the values that has either `null` or any boolean value are not being rendered. However it's not a problem on the Queries/Mutations sections: within the "Cached Data" tab, the json tree renders all the values correctly, with string values wrapped in quotes, and other values displayed as is.


alexTayanovsky:
hi @hwillson , is there anything else here I can help with? That's a pretty nasty regression, especially for those who rely on Apollo as a state management library

hwillson:
@alexTayanovsky sorry, I somehow missed this PR! I'll look at this shortly, and 🤞 get this published today. Thanks!

hwillson (approved):
Thanks very much @alexTayanovsky!

PR #659, opened 2021-08-11
Pt2: Live Subscriptions in Studio Explorer in Dev Tools

mayakoneval:
Previously in dev tools, subscription data was only reported when the button was pressed, even though we kept the socket open indefinitely. This PR shows data in realtime as we receive it, and closes the connection from the client when we receive a Subscription Termination Request from the embedded explorer.

Before:
https://user-images.githubusercontent.com/14367451/129082611-39a81f60-e756-478e-80d5-3c95effe5d25.mov

After:

https://user-images.githubusercontent.com/14367451/129082870-d5ee308a-ba3f-43a3-9bc1-65adbe8ddad0.mov

### DIFF
--- a/src/application/components/Explorer/Explorer.tsx
+++ b/src/application/components/Explorer/Explorer.tsx
@@ -17,10 +17,11 @@ import {
   receiveExplorerResponses,
   sendExplorerRequest,
   listenForResponse,
+  sendSubscriptionTerminationRequest,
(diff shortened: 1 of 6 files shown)


hwillson (approved):
This is awesome @mayakoneval - thanks very much!

(if you don't mind changing the base branch to `release-4.0` before merging, that would be 👍)

PR #661, opened 2021-08-12
Delete extra console.log

mayakoneval:
left this one in by accident!


hwillson (approved):
Great catch @mayakoneval - thanks!

PR #678, opened 2021-08-31
Some small fixes I made as I was iterating on Embeddable Explorer

mayakoneval:
renamed SUBCRIPTION_TERMINATION to EXPLORER_SUBSCRIPTION_TERMINATION to match the other constants & make some props optional, since they are optional when we send them from the embeddable explorer. Also updated the type of `variables` to match the embeddable explorer.


hwillson (approved):
Looks great, thanks @mayakoneval!

PR #686, opened 2021-09-08
Embedded Explorer communication naming changes

mayakoneval:
context: https://github.com/mdg-private/studio-ui/pull/4899

I tested this with that branch checked out locally, and everything seemed to work. 

TODO: figure out if we a post message query param


hwillson (approved):
Awesome, thanks @mayakoneval!

PR #704, opened 2021-09-22
README updates

mayakoneval:
I tried to update the readme as best as possible with what I have learned from working on this project. 

The last commit deals with the communication relay between the client and the devtools, which is pretty damn convoluted at the moment. I don't think its worth it to try and clean this up right now (and it works), so I just documented the path from what I understand currently. I don't know some of the answers to "why" this is setup the way it is, but I do know how it works to an extent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,11 @@ For builds, we use the `build` folder. After a build, all of the files needed to
 
 ### Debugging
 
-If there is an error in the devtools panel, you can inspect it just like you would inspect a normal webpage. Detach the inspector console from the window (if it's not already detached) by clicking the button with three vertical dots in the upper right corner of the console and selecting the detach option. With the detached console in focus, press `opt-cmd-I` again to open an inspector for the detached console (inspector inception). In this new inspector you will be able to inspect elements in the first inspector, including the devtools panel.
+If there is an error in the devtools panel, you can inspect it just like you would inspect a normal webpage.
+
+In Chrome, detach the inspector console from the window (if it's not already detached) by clicking the button with three vertical dots in the upper right corner of the console and selecting the detach option. With the detached console in focus, press `opt-cmd-I` again to open an inspector for the detached console (inspector inception). In this new inspector you will be able to inspect elements in the first inspector, including the devtools panel.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,9 +113,35 @@ The devtools folder structure mirrors this architecture. The source code for the
...
+
+**`explorerRelay.ts`** is a file with a bunch of exported functions that are specific to the Explorer network communications for executed operations. `devtools.ts` uses the functions as callbacks for its incoming messages, and `Explorer.tsx` uses the functions to dispatch network requests & accept responses to display in the embedded Explorer.
+
+#### Explorer network communication
+
+When requests are trigger by the user from Explorer, `sendExplorerRequest` in `explorerRelay.ts` dispatches an `EXPLORER_REQUEST` event which is picked up in `devtools.ts` and forwarded to the client. In `hook.ts` the `EXPLORER_REQUEST` message is listened for, and an operation is executed. When a response for this network request is recieved by the client, `EXPLORER_RESPONSE` is sent to the tab by the client in `hook.ts`. This message is forwarded in `tabRelay.ts` to the devtools, which calls `sendResponseToExplorer` which is picked up by `receiveExplorerResponses` called in `Explorer.tsx`.
```
hwillson, line 136: "When requests are trigger ..." -> "When requests are triggered ..."

hwillson (approved):
This is awesome @mayakoneval, thanks! Just one small typo. Thanks!

README.md
mayakoneval, line 136:
```suggestion
When requests are triggered by the user from Explorer, `sendExplorerRequest` in `explorerRelay.ts` dispatches an `EXPLORER_REQUEST` event which is picked up in `devtools.ts` and forwarded to the client. In `hook.ts` the `EXPLORER_REQUEST` message is listened for, and an operation is executed. When a response for this network request is recieved by the client, `EXPLORER_RESPONSE` is sent to the tab by the client in `hook.ts`. This message is forwarded in `tabRelay.ts` to the devtools, which calls `sendResponseToExplorer` which is picked up by `receiveExplorerResponses` called in `Explorer.tsx`.
```

PR #713, opened 2021-09-29
Query param renames

mayakoneval:
We just changed the names of these this week :). `shouldPersistState` is whether or not to keep state in local storage. Do we want this to be true or false for dev tools? Do folks expect a clean explorer when they come to dev tools?

### DIFF
--- a/src/application/components/Explorer/Explorer.tsx
+++ b/src/application/components/Explorer/Explorer.tsx
@@ -255,7 +255,7 @@ export const Explorer = ({ navigationProps, embeddedExplorerProps }: {
           </label>
         </FullWidthLayout.Header>
         <FullWidthLayout.Main css={mainStyles}>
-          <iframe id="embedded-explorer" css={iFrameStyles} src={`${EMBEDDABLE_EXPLORER_URL}?postMessageOperations=true&showHeadersAndEnvVars=false&theme=${color}`}/>
+          <iframe id="embedded-explorer" css={iFrameStyles} src={`${EMBEDDABLE_EXPLORER_URL}?sendRequestsFrom=parent&shouldPersistState=true&showHeadersAndEnvVars=false&theme=${color}`}/>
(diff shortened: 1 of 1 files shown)


cheapsteak:
Is state shared between different pages or just on the page where it's opened?

If I open devtools on Medium's site, write something, open a new tab, go to Studio and open devtools, would the contents contain what I wrote in Medium's devtools tab?

mayakoneval:
@cheapsteak I should have checked this before, but it looks like the referrer is nothing from the chrome dev tools, so I will ship this as `shouldPersistState=false`, and if we want to include a (not public?) query param to set a localhost id, we could make it so this is persisted based on the current page.

src/application/components/Explorer/Explorer.tsx
```diff
@@ -255,7 +255,7 @@ export const Explorer = ({ navigationProps, embeddedExplorerProps }: {
           </label>
         </FullWidthLayout.Header>
         <FullWidthLayout.Main css={mainStyles}>
-          <iframe id="embedded-explorer" css={iFrameStyles} src={`${EMBEDDABLE_EXPLORER_URL}?postMessageOperations=true&showHeadersAndEnvVars=false&theme=${color}`}/>
+          <iframe id="embedded-explorer" css={iFrameStyles} src={`${EMBEDDABLE_EXPLORER_URL}?sendRequestsFrom=parent&shouldPersistState=true&showHeadersAndEnvVars=false&theme=${color}`}/>
```
mayakoneval, line 258:
```suggestion
          <iframe id="embedded-explorer" css={iFrameStyles} src={`${EMBEDDABLE_EXPLORER_URL}?sendRequestsFrom=parent&shouldPersistState=false&showHeadersAndEnvVars=false&theme=${color}`}/>
```

cheapsteak (approved):
looks good!

PR #734, opened 2021-10-18
fix: do not break Import Maps

### LINKED ISSUE #464: Breaks Support for Import Maps
**Intended outcome:**
It should be possible to use apollo client devtools on websites that use import maps.

**Actual outcome:**
Because Apollo Client Devtools injects ES modules into the page, this causes all subsequent import maps in the page to not be processed, giving the import map error - `An import map is added after module script load was triggered.`

I've posted more information about this in https://github.com/WICG/import-maps/issues/248.

For example, try navigating to https://generator.jspm.io/ with the extension enabled in latest chrome with import maps support (89+). The error above comes up and the page will not load further unless the extension is disabled.

 - Browser: Chrome
 - Browser version: 89
 - Extension version: 3.02

hronro:
Fix #464

### DIFF
--- a/src/extension/tab/tab.ts
+++ b/src/extension/tab/tab.ts
@@ -10,11 +10,13 @@ if (typeof document === "object" && document instanceof HTMLDocument) {
   const script = document.createElement("script");
   script.setAttribute("type", "module");
   script.setAttribute("src", chrome.extension.getURL("hook.js"));
-  const importMap = document.querySelector("script[type=\"importmap\"]");
-  if (importMap != null) {
-    importMap.parentNode?.insertBefore(script, importMap.nextSibling);
-  } else {
-    const head = document.head || document.getElementsByTagName("head")[0] || document.documentElement;
-    head.insertBefore(script, head.lastChild);
-  }
+  document.addEventListener("DOMContentLoaded", () => {
+    const importMap = document.querySelector("script[type=\"importmap\"]");
+    if (importMap != null) {
+      importMap.parentNode?.insertBefore(script, importMap.nextSibling);
+    } else {
+      const head = document.head || document.getElementsByTagName("head")[0] || document.documentElement;
+      head.insertBefore(script, head.lastChild);
+    }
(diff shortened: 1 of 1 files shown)


MrDoomBringer:
Hey @hronro, thanks a ton for the PR!

I went ahead and added a commit that waits for the `DOMContentLoaded` event before injecting the script, just to be sure we have access to the importmap script if it's available. Tested on https://generator.jspm.io/, the AC [fullstack tutorial](https://github.com/apollographql/fullstack-tutorial) (w/ `apollo/client@3.6.9`), and the [DevTools example app](https://github.com/apollographql/apollo-client-devtools/tree/main/development) (`apollo/client@3.7.1`)

Will get this merged shortly pending a quick once-over

All the best,
Emmanuel :-)

bignimbus (approved):

src/extension/tab/tab.ts
```diff
@@ -10,6 +10,13 @@ if (typeof document === "object" && document instanceof HTMLDocument) {
...
   script.setAttribute("src", chrome.extension.getURL("hook.js"));
-  const head = document.head || document.getElementsByTagName("head")[0] || document.documentElement;
-  head.insertBefore(script, head.lastChild);
+  document.addEventListener("DOMContentLoaded", () => {
+    const importMap = document.querySelector("script[type=\"importmap\"]");
+    if (importMap != null) {
+      importMap.parentNode?.insertBefore(script, importMap.nextSibling);
+    } else {
+      const head = document.head || document.getElementsByTagName("head")[0] || document.documentElement;
+      head.insertBefore(script, head.lastChild);
```
guybedford, line 19: This branching shouldn't actually even be necessary, since a dynamic load at DOM Content Loaded time would be deemed to always load after the import map so wouldn't interfere with the import map.

If wanting to load sooner, another option could be to check if there is already a module script, and if so trust that it is safe to do an injection, otherwise waiting for DOM content loaded.

PR #794, opened 2022-01-08
NEBULA-613 "copy" button for variables should give valid JSON

SilverMaiden:
Switched from using `stringifyObject` to `JSON.stringify `so when a user copies variables, it's in JSON format.

[JIRA](https://apollographql.atlassian.net/browse/NEBULA-613)


cheapsteak (commented):
What are the changes in package-lock.json?

Strange to have changes there when there are no changes in package.json

cheapsteak (approved):
🎉

PR #827, opened 2022-03-09
fix(ui): fixes height of query, mutation, and cache content

benatshippabo:
## Fix

This fixes the height of the query content not taking the full height of the window.

### Before
image

### After
image

I believe this is because we forgot to set the height like we did in the here:
https://github.com/apollographql/apollo-client-devtools/blob/dc36a50dfdb82000401bfa6a0f6589069600c2d2/src/application/components/Layouts/FullWidthLayout.tsx#L27-L30

## Steps to reproduce bug
1. go to https://codesandbox.io/s/apollo-client-devtool-variables-forked-fuoro5
2. open apollo devtools and view one of the queries
3. notice it doesn't take the full height


benatshippabo:
Looks like it's been fixed

PR #828, opened 2022-03-09
feat: find apollo client instance on embedded iframes

boehlkers:
Per https://github.com/apollographql/apollo-client-devtools/discussions/380.

This PR adds discovery support for apollo client instances in same-origin iframes. It does so by looking through any embedded iframes and looking for window.__APOLLO_CLIENT__ instances therein.

I also added an iframe.html file to the client repo to test and validate the behavior; committing it may be overkill since it's not core functionality but willing to get feedback on that. Can verify by starting the local app with `npm run start:dev` and navigating to localhost:3000/iframe.html, which will load the demo app in an iframe.

### DIFF
--- a/development/client/public/iframe.html
+++ b/development/client/public/iframe.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title>Color App in an IFrame</title>
+</head>
+
+<body>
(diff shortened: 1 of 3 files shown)


jerelmiller:
Hey @boehlkers 👋 

Thanks for your patience! This PR is now quite outdated and we've implemented this instead in #1797. Thanks for your contribution!

PR #884, opened 2022-05-24
🐛 Firefox: set wider cross origin opener policy to enable auth popup from devtools

mayakoneval:
In chrome, we didn't need any special headers to allow the iframed embed to have an authorize button that triggers a  `window.open` action that has `window.opener` set. However, in Firefox, the `window.opener` is null because the iframe is opening a cross origin website (explorer.embed.apollographql.com opens studio.apollographql.com). 

I was testing with `embed.apollo.local` and `localhost:4000`, so maybe the subdomains are actually fine when we use the prod urls? Looking into it now

### DIFF
--- a/src/application/components/Explorer/GraphRefModal.tsx
+++ b/src/application/components/Explorer/GraphRefModal.tsx
@@ -48,7 +48,7 @@ export const GraphRefModal = ({
           height: 400,
           position: "absolute",
           margin: "auto",
-          top: "150px",
+          top: "20%",
           left: "calc(50% - 250px)",
           display: "flex",
           backgroundColor: "white",


src/application/components/Explorer/Explorer.tsx
```diff
@@ -85,12 +85,13 @@ const iFrameStyles = css`
...
 
 const authorizeButtonStyles = css`
-  width: 175px;
+  width: 185px;
   height: 25px;
   cursor: pointer;
   border: none;
   border-radius: 2px;
   color: ${colors.grey.darker};
+  font-size: 13px;
```
cheapsteak, line 94: double checking these CSS changes are intentional? 👀

cheapsteak (approved):
🎉

src/application/components/Explorer/Explorer.tsx
mayakoneval: yes! this was just something I saw when testing on firefox!

PR #885, opened 2022-05-24
✨ remove unnecessary COOP policy - tested with prod endpoints and we don't need it

mayakoneval:
looks like it might have just been with `https://embed.apollo.local:3000` specific to firefox


cheapsteak (approved):
ah nice

PR #886, opened 2022-05-24
add pop up callout to modal

mayakoneval:
Took me an hour to figure out why the auth flow wasn't working on firefox before releasing & its because you need to enable popups. Added that text to the modal so no user gets confused


cheapsteak (approved):
🙏 💭

PR #889, opened 2022-05-31
🐛 Improve reliability of devtools initializing 

Jephuff:
The Apollo client devtools tab is inconsistently showing up, after investigating it seems there are a few race conditions between initializing the devtools and finding the Apollo client. This PR fixes a few of those race conditions to hopefully fix or at least improve the reliability of the devtools showing up.

### DIFF
--- a/src/extension/devtools/devtools.ts
+++ b/src/extension/devtools/devtools.ts
@@ -41,8 +41,6 @@ function startRequestInterval(ms = 500) {
   return () => clearInterval(id);
 }
 
-sendMessageToClient(DEVTOOLS_INITIALIZED);
-
 let isPanelCreated = false;
 let isAppInitialized = false;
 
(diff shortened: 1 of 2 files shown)


cheapsteak:
oh whoops, I closed the jira ticket since it was a duplicate of the one Jeffrey's already working on, forgot that that would sync the PR to closed, my bad

mayakoneval (approved):
All these changes LGTM, I will say in checking this out and running locally, I still see the Apollo tab doesn't come up after a couple inspects on the same tab, but when you open a new tab it comes back. This was the same thing I was seeing before. 

Hard to tell if this is quantifiably better but definitely not worse! Wondering if you noticed any situations that were particularly better that I could verify with testing as well

Jephuff:
> 

hmm with these changes I couldn't get it to not show up :( 
so for the flow that makes it not show up for you is it this sort of flow?
load page -> open devtools (shows up) -> close devtools -> open devtools(shows up) -> close devtools -> open devtools(doesn't show up)

mayakoneval:
@Jephuff I just tested again and it works great, I don't know why it wasn't showing up before but this seems reliable now. I feel like this is good to go

PR #921, opened 2022-09-29
Mitigate duplicate active queries appearing in the Queries sidebar

MrDoomBringer:
This issue resolves `JIRA-NEBULA-743`

The Client Dev Tools would ometimes show duplicate active queries in the left side-panel. I believe this is because when getting queries it would manually read from the private `queryManager.queries` field, which explains the unexpected behavior we've been getting.

I've changed the logic to instead use the `queryManager.getObservableQueries()` function, which is supported and does exactly what we're trying to do here (i.e., get a list of active queries)
Before:
image

After:
image

### DIFF
--- a/src/extension/tab/broadcastQueries.js
+++ b/src/extension/tab/broadcastQueries.js
@@ -2,12 +2,9 @@
 // a Map. Devtools are expecting to work with an object, so this function
 // will convert an AC3 query info Map to an object, while also filtering out
 // the query details we don't need.
-function filterQueryInfo(queryInfoMap, useObservableQueries) {
+function filterQueryInfo(queryInfoMap) {
   const filteredQueryInfo = {};
   queryInfoMap.forEach((value, key) => {
-    if (useObservableQueries){
-      value = value.queryInfo;
-    }
     filteredQueryInfo[key] = {
       document: value.document,
       graphQLErrors: value.graphQLErrors,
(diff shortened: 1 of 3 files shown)


mayakoneval:
@MrDoomBringer I am getting `<whatever>.getObservableQueries is not a function` as a type error when opening the local colors app that comes with the repo & trying to use apollo client. When console logging the `hook.ApolloClient` object, I see `queries` but no `getObservableQueries`

alex-wilmer (dismissed):
This is definitely an improvement - we need to investigate duplicate queries in the `queryManager` as well - but this should be okay to merge

mayakoneval:
@MrDoomBringer just adding the context here too: 

> we discovered that getObservableQueries only exists on some AC version between 3.3 & 3.6 and up. How do yall feel about serving the slightly bad duplicate query version to folks on the older versions of AC?
Emmanuel will add a check to see if the function exists & use one or the other depending. This should fix :team-michelle: 's demo (edited) 

also - can we update the `broadcastQueries` function to share logic here 🙏 would be great if we could put it in one place

mayakoneval:
@MrDoomBringer also curious if folks on your team have historical context for `getObservableQueries` - can we confirm the version it was added in?

MrDoomBringer:
> @MrDoomBringer also curious if folks on your team have historical context for `getObservableQueries` - can we confirm the version it was added in?

Yup! This was added in https://github.com/apollographql/apollo-client/pull/7813, `@apollo/client@3.4.0-rc.4`

mayakoneval:
@MrDoomBringer ping me when you need a re-review please!

MrDoomBringer:
@mayakoneval Should be good for a re-review. Updated broadcastQueries.js to use client.getObservableQueries() as well, and added compatibility checks for client versions <` 3.4.0`

MrDoomBringer:
>  I am unable to test the broadcastQueries code at all, I wonder if we should just not change that file righ tnow?

I was having trouble testing that specific code too, I agree that it should be fine to leave it for now. Thanks!

PR #938, opened 2022-11-09
Fix CI tests failing

MrDoomBringer:
https://github.com/apollographql/apollo-client-devtools/pull/930 was causing some tests to fail because of a stylesheet import that wasn't completely explicit. This PR fixes that by specifying `/dist/style.css` rather than `/dist/style` in the import statements.


bignimbus (approved):
🚀

Issue #935, opened 2022-11-02
Unable to resize Queries, Mutations, and Cache tabs


cstoddart:
**Intended outcome:**

When moving my cursor over the edge of the pane in devtools I expect to be able to resize it.

**Actual outcome:**

There is no way to resize the panes within devtools.

**How to reproduce the issue:**

Open devtools, try to resize

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Browser version: 106
 - Extension version: 4.1.3

**Screenshot:**

<img width="534" alt="Screen Shot 2022-11-01 at 2 27 09 PM" src="https://user-images.githubusercontent.com/8518615/199511681-61a3c652-95c9-4aa2-ad27-d5233309ed79.png">

As you can see in the screenshot, the useful information in Cache is not easily viewable.

cstoddart:
Originally reported here: https://github.com/apollographql/apollo-client-devtools/issues/465

cstoddart:
Verified working as expected now. Thanks guys 🔥

PR #947, opened 2022-12-08
Makes the sidebar resizable

MrDoomBringer:
Fixes https://github.com/apollographql/apollo-client-devtools/issues/935

Allows the sidebar to be resized by dragging its edge.
image
The last step is to prevent that extra white area on the right from appearing when resizing the sidebar all the way to the left


alessbell:
Howdy @MrDoomBringer 👋 Hope things are well with you!

Hope you don't mind a couple of small fixes pushed to this branch and I'll get this shipped next week. Just noticed at large widths there's a layout issue:

CleanShot 2023-09-01 at 15 59 21

No action needed on your part (unless you want to, of course), looking forward to having this land :)

alessbell:
I ended up rewriting this with https://github.com/bvaughn/react-resizable-panels and I think the result has a couple of nice improvements :)

Since the default and min widths are percentages, it's usable even if you open it in "portrait" mode.

Before, can't access most of the devtools:
CleanShot 2023-09-05 at 17 37 47


After, every piece of UI is now onscreen, even if the query is overflowing a bit:
CleanShot 2023-09-05 at 17 37 50

And if you move the devtools to the bottom of the window, these are the same default width proportions as the last screenshot:
CleanShot 2023-09-05 at 17 37 59

CleanShot 2023-09-05 at 17 38 16

Just fixing up some tests and adding a changeset and will get this merged!

cstoddart:
Thank you for fixing this! 💪

PR #961, opened 2023-01-18
Upgrade web ext and use ES Module for eslint config

jerelmiller:
The existing version of `web-ext` was... less than ideal. While developing, any change to the extension would cause a new Chrome window to open. Not only this, but it was not properly opening to the right start URL.

Upgrading to the latest `web-ext` and `web-ext-plugin` fixed this, however the latest versions are written in ES modules. This PR also updates the Webpack config to use an ES Module format to allow use of the newest versions. Now the extension will reload the existing chrome window when changes are made.


webpack.config.mjs
```diff
@@ -37,7 +38,15 @@ module.exports = (env) => {
   ];
 
   if (env.NODE_ENV === "development") {
-    plugins.push(new WebExtPlugin({ target: env.TARGET }));
+    plugins.push(
+      new WebExtPlugin({ 
+        browserConsole: true,
+        runLint: false,
+        sourceDir: path.resolve(__dirname, 'build'),
+        startUrl: 'http://localhost:3000',
```
alessbell, line 46: Nit: we probably want to default to https since [insecure HTTP/2 is not supported even for localhost](https://web.dev/when-to-use-local-https/#when-to-use-https-for-local-development), to support the case where someone is debugging locally against a back-end using HTTP/2. (It should gracefully fallback to HTTP/1 but they may not realize/need to debug behavior specific to 2.)

```suggestion
        startUrl: 'https://localhost:3000',
```

alessbell (approved):
LGTM! Just left one comment 🚢

webpack.config.mjs
jerelmiller: I just tried this and unfortunately it doesn't open. This is what I see when I run `npm run chrome` for example.

Screenshot 2023-01-19 at 10 42 28 AM

Since this is configured to work with the app in the `development` directory, which runs on port 3000, I'm going to leave it as-is for now. Perhaps we could explore updating this in the future, but right now it looks like it will be a lot more changes to get https working correctly. FWIW, the development app has its own server and requests made to a remote resource are done from the server, so this might be less of an issue. We can always revisit though.

Is that ok? Or do you feel strongly we should try and get this working on https?

jerelmiller: Oh I see what you're looking at. Something that might not be apparent in this PR is this line: https://github.com/apollographql/apollo-client-devtools/blob/upgrade-web-ext/webpack.config.mjs#L40

```tsx
if (env.NODE_ENV === "development") {
```

This Webpack plugin is a development-only tool that works with [`web-ext`](https://github.com/mozilla/web-ext) to, as it says, "help build, run, and test web extensions". This particular webpack plugin provides a helpful wrapper around `web-ext` that does a couple things:

Automatically spins up a browser instance with the local version of the extension installed onto it. Refreshes the extension for you when as you develop so that you can observe changes to the extension as you make them. Much faster than stopping/starting the tool each time you make a change.

This won't have anything to do with the actual published browser extension. This is only for development purposes to allow you to develop the extension locally.

Hope that helps!

jerelmiller: Oops, meant to include this in the last comment. 

This `startUrl` property just tells the `web-ext` what the URL should be when the browser spins up in development. It's a small step that saves you from having to type `localhost:3000` in the address bar when you run `npm run chrome` or `npm run firefox`.

PR #993, opened 2023-03-05
Use `browser` namespace with `webextension-polyfill` instead of `chrome` namespace

jerelmiller:
`browser` is the [proposed standard](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Build_a_cross_browser_extension) for building cross-browser extensions, which also supports the use of promises instead of callbacks for async APIs. Our extension currently uses the `chrome` namespace for interaction in our extension. While this has always worked in both Firefox and Chrome, `browser` will be the future supported API. 

This PR updates the use of `chrome` APIs to use `browser` by installing the [`webextension-polyfill`](https://github.com/mozilla/webextension-polyfill) to provide compatibility with Chrome.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,6 +58,7 @@
         "typescript": "4.9.5",
         "web-ext": "7.5.0",
         "web-ext-plugin": "2.7.0",
+        "webextension-polyfill": "^0.10.0",
         "webpack": "5.75.0",
         "webpack-cli": "4.10.0"
       }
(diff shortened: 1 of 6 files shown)


phryneas (approved):
Looks good to me. Always good to move some callbacks over to promise-based APIs! 👍

alessbell (approved):
Nice one! This is a great improvement, LGTM 🎉

PR #994, opened 2023-03-05
Modernize many of the packages and upgrade the code theme to something more Apollo-y

jerelmiller:
This PR upgrades to the latest versions of many packages including React and Apollo. As part of this upgrade, I discovered a couple packages that were not listed as compatible with React 18, so I decided to replace those packages with similar packages that are more actively maintained.

These libraries are:
* `@reach/tabs` --> `@radix-ui/react-tabs`
* `react-graphql-syntax-highlighter` --> `prism-react-renderer`

While working on the switch to `prism-react-renderer` I also brought over the code theme used in our docs, which fits much more nicely with the rest of the the devtools theme. 

**Before**
Screenshot 2023-03-05 at 4 34 41 PM

**After**
Screenshot 2023-03-05 at 4 34 50 PM

---

NOTE: Merge #993 first as it contains changes also found in this PR


src/application/components/SyntaxHighlighter.tsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
```
phryneas, line 1: I think this can go?

src/application/components/SyntaxHighlighter.tsx
```diff
@@ -0,0 +1,132 @@
...
+      theme={theme}
+      code={code}
+    >
+      {({ className, style, tokens, getLineProps, getTokenProps }) => {
+        return (
+          <pre 
+            className={[className, outerClassName].filter(Boolean).join(' ')} 
+            style={style}
+          >
+            <code>
```
phryneas, line 114: Minor nitpick: maybe add the className & style from above directly on the `code` (with `white-space: pre;` and `word-break: keep-all;`) and omit the `pre` element?

src/extension/devtools/devtools.ts
```diff
@@ -48,94 +49,93 @@ devtools.addConnection(EXPLORER_SUBSCRIPTION_TERMINATION, () => {
...
-
-          if (!isAppInitialized) {
-            initialize();
-            writeData({ queries, mutations, cache: JSON.stringify(cache) });
-            isAppInitialized = true;
-          }
+    );
 
-          clearRequestInterval = startRequestInterval();
+    isPanelCreated = true;
```
phryneas, line 60: I would even move the `isPanelCreated = true;` before the asynchronous call to `browser.devtools.panels.create` so we don't get into any race condition where this can be called multiple times while the panel is being created, resulting in multiple panels (probably impossible or it would already have happened before, but better be sure).

src/extension/devtools/devtools.ts
```diff
@@ -48,94 +49,93 @@ devtools.addConnection(EXPLORER_SUBSCRIPTION_TERMINATION, () => {
   sendMessageToClient(EXPLORER_SUBSCRIPTION_TERMINATION);
 });
 
-devtools.listen(CREATE_DEVTOOLS_PANEL, ({ payload }) => {
+devtools.listen(CREATE_DEVTOOLS_PANEL, async ({ payload }) => {
   if (!isPanelCreated) {
```
phryneas, line 53: Maybe, while we're at it, inverse this and add an early `return`?
```suggestion
  if (isPanelCreated) return;
```

tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "skipLibCheck": true,
     "module": "commonjs",
     "target": "es6",
-    "jsx": "react",
+    "jsx": "react-jsx",
+    "jsxImportSource": "@emotion/react",
     "resolveJsonModule": true
```
phryneas, line 12: This tsconfig is making me itchy. Once this is merged, I'd love to turn on `strict: true` and go over the whole repo with a broom 😅
Any objections?

src/application/components/Layouts/__tests__/Navigation.test.tsx
```diff
@@ -19,19 +19,22 @@ describe("<Navigation />", () => {
...
-    const { container } = renderWithApolloClient(<Navigation {...props} />);
+  it("can select a navigation item", async () => {
+    const { container, user } = renderWithApolloClient(<Navigation {...props} />);
     const buttons = container.querySelectorAll("button");
     const lastButton = buttons[buttons.length - 1];
     expect(lastButton).not.toHaveStyleRule("color", colors.silver.lighter);
-    user.click(lastButton);
+    await act(async () => {
+      await user.click(lastButton);
+    });
```
phryneas, line 37:
```suggestion
    await act(() => user.click(lastButton));
```
This already returns a promise, no need to double-await :)

src/application/components/Mutations/__tests__/Mutations.test.tsx
```diff
@@ -73,15 +72,17 @@ describe("<Mutations />", () => {
...
-    const header = getByTestId("header");
+    const header = screen.getByTestId("header");
     expect(within(header).getByText("Unnamed")).toBeInTheDocument();
 
-    const sidebar = getByTestId("sidebar");
-    user.click(within(sidebar).getByText("AddColorToFavorites"));
+    const sidebar = screen.getByTestId("sidebar");
+    await act(async () => {
+      await user.click(within(sidebar).getByText("AddColorToFavorites"));
+    });
```
phryneas, line 85:
```suggestion
    await act(() => user.click(within(sidebar).getByText("AddColorToFavorites")));
```

phryneas:
All in all, only a few minor nitpicks - great job keeping the DevTools up-to-date!

src/application/components/Queries/QueryViewer.tsx
```diff
@@ -53,19 +47,30 @@ export const queryStringHeader = css`
...
   grid-area: queryDataHeader;
   height: 100%;
   background-color: transparent;
   ${headerStyles}
   color: ${colors.grey.lighter};
 
   button {
-    padding-top: 0;
-    padding-bottom: 1.7rem;
+    padding: 0 0.5em 1.7rem;
```
alessbell, line 58: Nit: mixing `em` and `rem`. As an aside, I'd be in favor of migrating to Tailwind or a similar utility class-based system at some point in the future :)

alessbell (approved):
Some really nice improvements here! Re-using existing code theming variables also means the color contrast now passes the WCAG AA rating, an important a11y win 🎉

tsconfig.json
jerelmiller: @phryneas 100%. I was also bothered by this. So many places where types are just plain missing (especially component props), so I'm not opposed at all. I avoided doing it in this PR since it was already getting a bit lengthy with all the upgrades to the packages, but I'd love for that improvement.

src/application/components/Queries/QueryViewer.tsx
jerelmiller: Ah ya good callout.

My goal with this PR wasn't so much to fix CSS issues (like this), but rather ensure compatibility with the existing CSS. This used to use [`@reach/tabs`](https://reach.tech/tabs) which included its own [styling](https://github.com/reach/reach-ui/blob/dev/packages/tabs/styles.css). The `0.5em` value came from the [`@reach/tabs` stylesheet](https://github.com/reach/reach-ui/blob/43f450db7bcb25a743121fe31355f2294065a049/packages/tabs/styles.css#L21), which I blindly copied over 😄.

I might leave this for now and would love to come back through this with a fine-tooth comb later on. I'm sure there are many other areas like this that could use improvement.

src/application/components/SyntaxHighlighter.tsx
jerelmiller: `<pre><code>` seems to be the [recommended way](https://www.codiga.io/blog/display-code-snippets-in-html/) to represent a block of code in HTML. Would you be ok leaving this as is?

src/application/components/Layouts/__tests__/Navigation.test.tsx
jerelmiller: Good catch! Updated in [0f71fc5](https://github.com/apollographql/apollo-client-devtools/pull/994/commits/0f71fc56f4b281ccb604b6ce7e42a97c93246b66)

src/application/components/Mutations/__tests__/Mutations.test.tsx
jerelmiller: [0f71fc5](https://github.com/apollographql/apollo-client-devtools/pull/994/commits/0f71fc56f4b281ccb604b6ce7e42a97c93246b66)

src/extension/devtools/devtools.ts
jerelmiller: Good call! Updated in [fc06c93](https://github.com/apollographql/apollo-client-devtools/pull/994/commits/fc06c930bef35564166b0fd239f2d630c56a9ccb)

src/extension/devtools/devtools.ts
jerelmiller: I might leave this for now only because its worked like this for a while. I would absolutely love to come back to some of this logic anyways and perhaps we could do a refactor of this code a bit more. Totally agree though that its possible this could create race conditions.

Perhaps we could put this in a separate PR that calls this out a bit better?

src/application/components/SyntaxHighlighter.tsx
phryneas: I don't have any strong feelings here :)

src/application/components/SyntaxHighlighter.tsx
jerelmiller: Good catch! Updated in [4b72ba3](https://github.com/apollographql/apollo-client-devtools/pull/994/commits/4b72ba3279c53f4754537a17858541a41efe17aa)

PR #1037, opened 2023-07-28
Make message type more clearer by adding a prefix

jhen0409:
Related https://github.com/apollographql/apollo-client-devtools/discussions/1036

This allows the message to be better identify/filter if others are also using `window.addEventListener('message', )` in the page.

Tested on https://github.com/jhen0409/react-native-debugger/pull/783, currently using a patch.


jerelmiller:
Hey @jhen0409 👋 

This will be superseded by https://github.com/apollographql/apollo-client-devtools/pull/1226 which aims to rework much of the message passing, including an identifier as to the source of the message from anything published by the extension. As such I'm going to go ahead and close this. Thanks for raising this issue!

PR #1052, opened 2023-09-01
Improve type safety with `noImplicitAny`

jerelmiller:
Partially addresses #1047

Enables the `noImplicitAny` TS rule and fixes all the TS errors as a result. In addition to the fixes, I provided a few improvements:

- Added codegen for the local schema to improve type safety of client-only queries
- Added [`ts-reset`](https://www.totaltypescript.com/ts-reset) to fix some edge case weirdness
- Some minor refactoring on `Queries` and `Mutations` views that avoids the need for 2 queries and instead uses a plain `array.find(...)` to get the selected query.

PR #1056, opened 2023-09-01
Improve cache search

jerelmiller:
Searching the cache in the devtools has always felt a little bit broken. At present, it doesn't do any filtering of the list, so matches against the search string are very difficult to find if the list of cache keys is large (see example in [this discussion](https://github.com/apollographql/apollo-client-devtools/discussions/867)). Furthermore, it would match against values in the cache entry itself, but this was difficult to find since it required you to activate the cache entry to see it.

This PR improves the search functionality by matching only against cache keys and actively filtering that list as the user types. The matched string in the cache key is also now highlighted so its obvious how the record is filtered. This should improve cases where you might have an ID and want to quickly see what what a record with that ID contains by pasting that ID into the search field.

**Before**
search-old

**After**
search-new

### DIFF
--- a/.changeset/eight-ants-decide.md
+++ b/.changeset/eight-ants-decide.md
@@ -1,0 +1,5 @@
+---
+"apollo-client-devtools": patch
+---
+
+Improve searching the cache by filtering the list of cache ids that match the search term and highlight the matched substring. This change removes the matching against the cache values as that was difficult to determine why a match occurred.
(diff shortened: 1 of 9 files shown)


alessbell (approved):
This is such a great usability improvement! 🚀

jerelmiller:
Will merge once #1052 is merged since this branches off of that 🙂

alessbell:
Just noticed that :) Was hoping I could sneak this into the next release I'm about to cut but can't review the other one before signing off today 😅

PR #1057, opened 2023-09-02
Begin migrating to Tailwind

jerelmiller:
Leaving this open with minimal changes to see if this is something we want to explore. After having worked with Tailwind.css for the Spotify Showcase, I find it to be a great way to add styles without having to touch a bunch of files to do so.

The current solution uses [Emotion](https://emotion.sh/docs/introduction) for styling. A few reasons why I'd like to consider migrating to tailwind:
- Its static CSS which can be a bit better for performance than CSS-in-JS solutions (though likely that doesn't matter very much in the devtools)
- Great tooling support
- Reduces some noise around the current style setup
  - A lot of components have their own style declarations in a variable elsewhere in the file, which leads to longer files
  - Styles are randomly imported/exported in various areas.

The biggest downside that others state are the much noisier `className` props. It can be a tad annoying when you have to apply LOTS of styles to it, but overall I don't find this much to be a nuisance. In fact, its great to be able to tweak a style by throwing another class in the list. Also much easier to visualize what something is going to look like just by looking at the component.

Figured I'd put this PR out there to figure out if this is something we want to do. If we like the idea of migrating, I'll spend some more time getting the rest of the styles changed. For now, this migrates the `Cache.tsx` component over to tailwind to get an idea of the differences in solutions.

I will say, I have used and love Emotion. If we decide to stick with it, I'm not opposed. Figured I'd toss this discussion point in there to see how we all think about it.


src/application/components/Cache/Cache.tsx
```diff
@@ -12,30 +9,6 @@ import { Loading } from "./common/Loading";
...
-
-const headerLabelStyles = css`
-  margin: ${rem(3)} 0 0 ${rem(8)};
-  font-family: "Source Sans Pro", sans-serif;
-  color: ${colors.grey.light};
-  text-transform: uppercase;
-  font-size: ${rem(11)};
-`;
-
-const noDataStyles = css`
```
jerelmiller, line 29: To me, this is the biggest advantage of this approach. These `css` blocks take up a lot of vertical space and are pretty disconnected from the component they style, which means a lot of back and forth to figure out the styles applied the element.

src/application/components/Cache/Cache.tsx
```diff
@@ -78,15 +51,21 @@ export function Cache({
             />
           </Fragment>
         ) : (
-          <h3 css={noDataStyles}>No cache data</h3>
+          <h3 className="ml-3 uppercase text-sm font-normal pt-4 text-white/50 tracking-wider">
```
jerelmiller, line 54: On the flipside, as more styles are needed, this can grow to be a bit large, which is the most commonly stated downside of Tailwind 😆

jerelmiller:
@phryneas @alessbell tagging you for review to start a discussion in here to see if this is something we want to entertain, or leave as-is.

alessbell (approved):
+1. I think merging this PR as is and doing an incremental refactor over time is the way to go - I could have used this in the resizeable sidebar PR :) Thanks for suggesting this!

alessbell:
Any other thoughts here @phryneas @jerelmiller? If not I can resolve the conflicts and get this merged 👍

jerelmiller:
@phryneas I know you had some thoughts about this and would love if you could summarize here.

jerelmiller:
Decision has been made to adopt this moving forward. Opening this PR out of draft.

@alessbell feel free to take this and fix the conflicts/merge it when you get the chance 🙂

alessbell:
Woohoo - happy with where this is at so I'm going to go ahead and merge, @jerelmiller 🎉 Created https://github.com/apollographql/apollo-client-devtools/issues/1094 to track the rest of the work.

Just tested the packed extension in Chrome/FF, will attempt an automated deploy once I merge the version packages PR :)

CleanShot 2023-09-12 at 17 40 49

Issue #1045, opened 2023-08-30
Remove apollo-link-state support?


alessbell:
While doing some maintenance, linting the codebase turned up the fact that `key` here is undefined: https://github.com/apollographql/apollo-client-devtools/blob/dc1547004ef20dc5af407b951340501aae06b753/src/extension/tab/links.js#L132

It looks like support for `apollo-link-state` broke in [January 2021](https://github.com/apollographql/apollo-client-devtools/commit/67cc212f51d662149eedee4bf6e04cf2094ec7e9).

Since `apollo-link-state` was itself deprecated in 2019, and seeing no bug reports, it seems like it's time to remove this feature.

PR #1070, opened 2023-09-05
Remove apollo link state support

alessbell:
Closes https://github.com/apollographql/apollo-client-devtools/issues/1045.

It looks like support for `apollo-link-state` broke in January 2021. It has long been deprecated (2019 or 2020). We could bump the minor version here if we want to call this out more, but a new major seems like overkill since no one has reported it broken for the last two years :)


jerelmiller:
FWIW I'd be ok with a minor version here

jerelmiller (approved):
Good find! Great to remove tech debt where possible.

PR #1085, opened 2023-09-11
chore: fix broken scrolling

alessbell:
When introducing the resizable sidebar, scrolling broke. Fixed and tested in Chrome and FF.

CleanShot 2023-09-11 at 12 25 44
CleanShot 2023-09-11 at 12 26 58

### DIFF
--- a/src/extension/devtools/panel.html
+++ b/src/extension/devtools/panel.html
@@ -8,6 +8,9 @@
       body {
         overflow: hidden;
       }
+      ::-webkit-scrollbar {
+        display: none;
+      }
       #devtools {
         width: 100vw;
         height: 100vh;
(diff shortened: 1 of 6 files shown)


src/extension/devtools/panel.html
```diff
@@ -8,6 +8,9 @@
       body {
         overflow: hidden;
       }
+      ::-webkit-scrollbar {
```
alessbell, line 11: In dark mode, `overflow: scroll` means white bars in the reserved space for the scroll bars (in FireFox, it respects the mode the browser is in which is visually much nicer as you can see in the screenshot).

It's possible to style the webkit scroll bar, but the APIs between webkit/non-webkit browsers are different. Hiding the scrollbar in webkit browsers is the past of least CSS for now.

PR #1097, opened 2023-09-14
Remove unused files

alessbell:
While doing some work on the devtools I noticed these files hadn't been migrated to TS.

Taking a closer look, I realized the line `import { buildSchemasFromTypeDefs } from "./typeDefs";` in `src/extension/tab/links.js` was importing from a file that no longer existed 😅 No TS file = not a peep from our tooling.

None of their exports were used anywhere and the files weren't referenced/imported anywhere. The final test was validating the build output: I diffed every file before/after and there were no changes to the production extension, so they seem safe to remove.


jerelmiller (approved):
💯

PR #1128, opened 2023-10-05
Settings Modal

alessbell:
Adds a settings panel that appears as a modal:

CleanShot 2023-10-05 at 13 56 18

For now, it just displays the current version of the extension and links to the GitHub release notes for that version, but soon it will allow users to enable "preview" features, select light/dark mode directly and any number of future config settings.

### Notes

- When using `@headlessui/react`'s `Transition` component, I ran into [this TS error](https://github.com/tailwindlabs/headlessui/issues/2468) caused by the global config of `"jsxImportSource": "@emotion/react",`, so that is now configured on a per file basis wherever the `css` prop is used via `/** @jsxImportSource @emotion/react */`

- `SettingsModal` is a single use component. It seems like premature optimization to make a generic `Modal` since we may only have this one use case for the foreseeable future, and it can easily be extracted if needed.


package.json
```diff
@@ -37,6 +37,8 @@
     "@apollo/client": "^3.7.10",
     "@apollo/space-kit": "^9.11.0",
     "@emotion/react": "^11.11.1",
+    "@headlessui/react": "^1.7.17",
```
jerelmiller, line 40: Curious your thoughts on whether we should standardize around [Radix](https://www.radix-ui.com/) or [headless ui](https://headlessui.com/). We are using the `@radix-ui/react-tabs` component, but might be good to avoid the mix & match strategy for the "headless" components.

I don't have a strong opinion on direction or the other. I use Radix in the Spotify UI and I like it, but from the documentation on each they seem pretty similar. If we go this route, I'd ask that we migrate the tabs component over to `@headlessui/react` to keep it all in the same library.

src/application/components/Layouts/SettingsModal.tsx
```diff
@@ -0,0 +1,77 @@
+import { Dispatch, Fragment, MutableRefObject, SetStateAction } from "react";
+import { Dialog, Transition } from "@headlessui/react";
+
+declare const VERSION: string;
+
+export function Modal({
```
jerelmiller, line 6: [nit] Since this is the `SettingsModal.tsx` file, could you name this component to match? Seeing `<Modal />` used in the layout feels like I'm working with a primitive Modal component. Naming this `SettingsModal` makes it easy to understand that we are working with a specific modal implementation.

src/application/components/Layouts/SettingsModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+  open: boolean;
+  cancelButtonRef: MutableRefObject<null>;
+  setOpen: Dispatch<SetStateAction<boolean>>;
+}) {
+  return (
+    <Transition.Root show={open} as={Fragment}>
+      <Dialog
+        as="div"
+        className="relative z-10"
+        initialFocus={cancelButtonRef}
```
jerelmiller, line 20: I don't see this `cancelButtonRef` added to any other element. From what I see in the [docs](https://headlessui.com/react/dialog#managing-initial-focus) this should be attached to some kind of element (likely a `<button />`). Do you need this at all? Perhaps leftover from a refactoring?

src/application/components/Layouts/SettingsModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+declare const VERSION: string;
+
+export function Modal({
+  open,
+  setOpen,
+  cancelButtonRef,
+}: {
+  open: boolean;
+  cancelButtonRef: MutableRefObject<null>;
+  setOpen: Dispatch<SetStateAction<boolean>>;
```
jerelmiller, line 13: [nit] I tend to like standardizing on event callbacks from child -> parent with the `on<Event>` naming scheme to be able to pick out which are the event callbacks. I also prefer having the function signature a bit more generic so you aren't coupling it specifically to `useState`. It can be refactored much easier this way if we ever need to (e.g. if you needed to execute a side effect when closing the modal). 

```suggestion
  onOpen: (open: boolean) => void
```

`Dispatch<SetStateAction<boolean>>` should play just fine with this function signature, so passing `setOpen` directly to `onOpen` from the parent works.

```tsx
<Modal onOpen={setOpen} /> // works!
```

src/application/components/Layouts/SettingsModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) {
+  return (
+    <Transition.Root show={open} as={Fragment}>
+      <Dialog
+        as="div"
```
jerelmiller, line 18:
```suggestion
```

[nit] Looks like `div` is the default, so feel free to omit this.

src/application/components/Layouts/SettingsModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+              enter="ease-out duration-300"
+              enterFrom="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
+              enterTo="opacity-100 translate-y-0 sm:scale-100"
+              leave="ease-in duration-200"
+              leaveFrom="opacity-100 translate-y-0 sm:scale-100"
+              leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
+            >
+              <Dialog.Panel className="relative transform overflow-hidden rounded-lg bg-white px-4 pb-4 pt-5 text-left shadow-xl transition-all sm:my-8 sm:w-full max-w-sm sm:p-6">
+                <div>
+                  <div className="mt-1">
```
jerelmiller, line 48:
```suggestion
                  <div>
```

The modal content doesn't quite feel vertically centered and this `mt-1` class seems to be the culprit. Removing this makes the modal content feel a bit more balanced in the container.

**Before**

Screenshot 2023-10-05 at 12 30 19 PM


**After**
Screenshot 2023-10-05 at 12 27 06 PM

src/application/components/Layouts/SettingsModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+              as={Fragment}
+              enter="ease-out duration-300"
+              enterFrom="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
+              enterTo="opacity-100 translate-y-0 sm:scale-100"
+              leave="ease-in duration-200"
+              leaveFrom="opacity-100 translate-y-0 sm:scale-100"
+              leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
+            >
+              <Dialog.Panel className="relative transform overflow-hidden rounded-lg bg-white px-4 pb-4 pt-5 text-left shadow-xl transition-all sm:my-8 sm:w-full max-w-sm sm:p-6">
+                <div>
```
jerelmiller, line 47: I'm curious if this extra `div` is needed since you have another wrapping div as a child of this one. Could you consolidate into a single one?

In fact, the `Dialog.Panel` seems to do a pretty good job of wrapping the whole content, so I think you can remove both this `div` and the child `div` as well. This doesn't seem to change the content in any way if I remove both of these when testing locally.

jerelmiller (commented):
Had some minor nits for you, but otherwise looks like a great start! I can approve once you've had a chance to respond to a couple questions I have here. Thanks!

src/application/components/Layouts/SettingsModal.tsx
alessbell: Addressed in https://github.com/apollographql/apollo-client-devtools/pull/1128/commits/41bdf86d24adcb2decbd996401584e8c7914dc99

src/application/components/Layouts/SettingsModal.tsx
alessbell: Addressed in https://github.com/apollographql/apollo-client-devtools/pull/1128/commits/41bdf86d24adcb2decbd996401584e8c7914dc99

src/application/components/Layouts/SettingsModal.tsx
alessbell: Addressed in https://github.com/apollographql/apollo-client-devtools/pull/1128/commits/41bdf86d24adcb2decbd996401584e8c7914dc99

src/application/components/Layouts/SettingsModal.tsx
alessbell: Addressed in https://github.com/apollographql/apollo-client-devtools/pull/1128/commits/41bdf86d24adcb2decbd996401584e8c7914dc99

alessbell:
This should be ready for a final pass, @jerelmiller, thanks for the review!

Issue #1035, opened 2023-07-25
[bug] Devtools never appear if initial page does not contain an Apollo Client instance


alessbell:
**Intended outcome:**
When I navigate to a page that initializes Apollo Client in the browser, the devtools should appear even if Apollo Client is not detected in the browser when the initial page load occurs.

**Actual outcome:**
If you have an app that is purely server rendered initially, and navigate to a page that loads Apollo Client the devtools never appear. You need to refresh the browser from the page that loads Apollo on the client in order for devtools to appear.

**How to reproduce the issue:**
I reproduced this in the Next.js experimental repo's [polls-demo example](https://github.com/apollographql/apollo-client-nextjs/tree/main/examples/polls-demo).

If you start on the homepage, `/`, which is a React Server Component and navigate to the client component page, Apollo Client devtools never appear.

If you refresh the client component page, they do appear:

![CleanShot 2023-07-25 at 10 43 43](https://github.com/apollographql/apollo-client-devtools/assets/5139846/164892b3-230a-47ae-bf06-65d087b67593)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] **Mac OS**
 - Browser [e.g. chrome, safari] **Arc**
 - Browser version [e.g. 22] **1.0.1**
 - Extension version [e.g. 3.0] **4.1.4**

vezaynk:
My app has delayed initializing for the graphql client, and I can't get the devtools to work :(

jerelmiller:
I've tested this out with recent versions of the devtools and they are working much better with these environments. Since we now go into the "connected" state of the devtools when a client is registered, we no longer need to rely on a page refresh for this to work. As such I'm going to close this out as completed 🙂

PR #1220, opened 2024-02-07
Always show the devtools panel, regardless of whether we detect a client

jerelmiller:
There are no shortage of [reviews](https://chromewebstore.google.com/detail/apollo-client-devtools/jdkknkkbebbapilgoeccciglkfbmbnfm/reviews) on our extension page in the form of "does not work". I haven't nailed down exactly the reason for it, but the problem seems to be pretty intermittent and/or we have bugs with our communication between all the parts of the extension (also very likely).

Its very difficult for us to follow up with everyone that posts their issues in our review section because we cannot reply to those users as our own account, and we do not monitor the account that "owns" the extension. The reviews flood in, but we rarely get a GitHub issue with details that we can properly followup on to get an idea of what might be happening.

This PR seeks to address a new approach to try and move us in a positive direction. Rather than waiting until we detect a client instance to create a devtools panel (which might have bugs on its own), the panel is instead created immediately and visible anytime you open the devtools for any webpage. I've added additional status messages to communicate to the user what is happening (trying to connect, disconnected, etc.) to avoid guesswork on whether the devtools has actually connected to a client instance. This change means the user can at least see the devtools pane and interact with it, even if we aren't able to connect to a client instance. When the devtools is unable to connect, we provide suggestions to figure out why the client might not be connected properly. This makes it a bit easier to debug as a user why the devtools might not have been able to connect.

While this PR likely doesn't address the underlying issues plaguing these users, I'm hoping this change means we'll at least get more GitHub issues opened that we can followup on. At the very least, we are able to show the user SOMETHING that indicates the extension is working and is actually trying to connect, rather than it just feeling totally broken because the panel never gets created.

### Screenshots

**Loading**
Screenshot 2024-02-07 at 9 12 03 PM

**Connected**
Screenshot 2024-02-07 at 9 12 49 PM

**Disconnected**
Screenshot 2024-02-07 at 9 13 13 PM

**Unable to communicate with tab**
Screenshot 2024-02-07 at 9 22 12 PM

**Client not found**
Screenshot 2024-02-07 at 9 12 08 PM
Screenshot 2024-02-07 at 9 12 21 PM


package.json
```diff
@@ -39,8 +39,10 @@
     "@apollo/icons": "^0.0.8",
     "@apollo/tailwind-preset": "^0.1.3",
     "@headlessui/react": "^1.7.17",
+    "@headlessui/tailwindcss": "^0.2.0",
     "@radix-ui/react-tabs": "^1.0.2",
     "clsx": "^2.1.0",
+    "framer-motion": "^11.0.3",
```
jerelmiller, line 45: Framer motion seems to be the only one that I can get proper [exit animations](https://www.framer.com/motion/animate-presence/). I could not get this using the built-in [`Transition` component](https://headlessui.com/react/transition) from [headless UI](https://headlessui.com/).

src/application/components/ClientNotFoundModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+          discovered. This is most commonly fixed by setting the{" "}
+          <ConnectToDevToolsOptionLink /> to <code>true</code>.
+        </p>
+        <p className="mt-4">
+          If this was reached in error, please try one of the following
+          suggestions or click the &quot;Retry&quot; button to try looking for
+          the client again.
+        </p>
+        <div className="mt-4 flex flex-col gap-2">
+          <Disclosure>
```
jerelmiller, line 48: Please review each of these for accuracy and helpfulness! If you have any suggestions I missed for this first round, let me know and I'd be happy to add them!

src/extension/constants.ts
```diff
@@ -1,9 +1,13 @@
 export const REQUEST_TAB_ID = "request-tab-id";
-export const DEVTOOLS_INITIALIZED = "devtools-initialized";
-export const CREATE_DEVTOOLS_PANEL = "create-devtools-panel";
+export const CLIENT_NOT_FOUND = "client-not-found";
```
jerelmiller, line 2: I did not choose to prefix these quite yet. There is more I'd like to investigate with our communication between the various areas of our extension, but I'd like to do that all at once instead of combining that in this PR.

src/application/App.tsx
```diff
@@ -36,97 +84,125 @@ const GET_OPERATION_COUNTS: TypedDocumentNode<
...
   const selected = useReactiveVar<Screens>(currentScreen);
-  const reloading = useReactiveVar<boolean>(reloadStatus);
+  const state = useReactiveVar(devtoolsState);
   const [embeddedExplorerIFrame, setEmbeddedExplorerIFrame] =
     useState<HTMLIFrameElement | null>(null);
 
-  // During a reload, reset the current screen to Queries.
   useEffect(() => {
-    if (reloading && selected !== Screens.Queries) {
-      currentScreen(Screens.Queries);
```
jerelmiller, line 49: I got rid of this weird reload behavior where it would always navigate you back to the queries screen anytime the page was reloaded. I couldn't think of a good reason why we couldn't just leave the user on the same screen they were looking at.

src/application/App.tsx
```diff
@@ -36,97 +84,125 @@ const GET_OPERATION_COUNTS: TypedDocumentNode<
...
+    let timeout: NodeJS.Timeout;
+    // Don't show connected message on the first render if we are already
+    // connected to the client.
+    if (!mountedRef.current && state === "connected") {
+      return;
     }
-  }, [reloading, selected]);
 
-  if (reloading) {
-    return <div></div>;
```
jerelmiller, line 54: I also discovered that it would empty out the screen anytime a page reload happens. I opted to remove this and keep the existing UI up while the page is being reloaded. The user will see a "disconnected" message in the bottom instead until the page has loaded and the client instance is detected again.

src/application/App.tsx
```diff
@@ -36,97 +84,125 @@ const GET_OPERATION_COUNTS: TypedDocumentNode<
...
-  }
+    if (state === "notFound") {
+      setClientNotFoundModalOpen(true);
+    }
+
+    const dismiss = BannerAlert.show(ALERT_CONFIGS[state]);
```
jerelmiller, line 108: I'm opting to show a banner for the `notFound` state alongside the modal, that way if the user closes the modal, there is still a way for them to retry the connection.

src/application/index.tsx
```diff
@@ -182,15 +182,6 @@ export const writeData = ({
   cacheVar(cache);
 };
 
-export const handleReload = () => {
-  reloadStatus(true);
-};
-
-export const handleReloadComplete = () => {
-  reloadStatus(false);
-  client.resetStore();
```
jerelmiller, line 191: This behavior is still maintained, but has been moved over to the [panel](https://github.com/apollographql/apollo-client-devtools/pull/1220/files#diff-7c1d36e9e85ad36384132856e68b4e2eaf0d11d4d64e911381c8878329a5f38aR40-R42) code. We reset the queries/mutations/cache anytime we connect to the client.

src/extension/devtools/devtools.ts
```diff
@@ -38,71 +111,101 @@ function sendMessageToClient(message: string) {
...
+const unsubscribers = new Set<() => void>();
+
+function unsubscribeFromAll() {
+  unsubscribers.forEach((unsubscribe) => unsubscribe());
+  unsubscribers.clear();
+}
 
+let connectedToPanel = false;
+
+async function createDevtoolsPanel() {
```
jerelmiller, line 137: This is the biggest change in the whole PR. Rather than waiting for a `CREATE_DEVTOOLS_PANEL` message, we unconditionally create it. See the PR description for why I believe this is a good idea 🙂

src/extension/devtools/devtools.ts
```diff
@@ -122,29 +225,20 @@ devtools.listen<string>(CREATE_DEVTOOLS_PANEL, async ({ payload }) => {
       `background:tab-${inspectedTabId}:client`
     );
 
-    // Listen for tab reload from background
-    removeReloadListener = devtools.listen(RELOADING_TAB, () => {
```
jerelmiller, line 126: The notion of "tab reloading" has gone away in favor of thinking about this as a "disconnect" event. The user may navigate away from the current application rather than reloading the page which would fire the same event.

src/extension/devtools/devtools.ts
```diff
@@ -122,29 +225,20 @@ devtools.listen<string>(CREATE_DEVTOOLS_PANEL, async ({ payload }) => {
...
-        clearRequestInterval = startRequestInterval();
-        handleReloadComplete();
-        removeListener();
-      });
-    });
+    panelHidden = false;
   });
 
   panel.onHidden.addListener(() => {
-    isPanelCreated = false;
```
jerelmiller, line 139: Hiding the panel happens when the user switches tabs in the devtools, not just closing them altogether. The notion that the panel was no longer created here before was incorrect.

src/application/components/ClientNotFoundModal.tsx
alessbell: These are great! What a big improvement to the troubleshooting UX 😄

src/application/components/ClientNotFoundModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                that you can use to check if the client was discovered. Check
+                your browser console for the{" "}
+                <code>window.__APOLLO_DEVTOOLS_GLOBAL_HOOK__.ApolloClient</code>{" "}
+                variable to see if it references your Apollo Client instance. If
+                this value is <code>undefined</code>, this likely indicates your
+                client was not discovered in time. Click the &quot;Retry&quot;
+                button below to try connecting to your Apollo Client instance
+                again.
+              </p>
+              <p className="mt-4">
```
alessbell, line 168: The only thing I'd add here is this also occurs when using our Next.js experimental package if the first page(s) you load don't render any client components that initialize AC in the first 10 seconds: https://github.com/apollographql/apollo-client-devtools/issues/1035

We may want to mention that case specifically since you can now just click reload after navigating, and close that issue 🎉

jerelmiller: Thats a good callout! I think one more "I'm using Next.js" disclosure is probably a good idea 🙂

src/application/stateMachine.ts
```diff
@@ -0,0 +1,135 @@
...
+  onTransition: (state: State, listener: () => void) => () => void;
+  matches: (state: State) => boolean;
+}
+
+type MachineConfig<
+  State extends string,
+  EventName extends string,
+  Context extends Record<string, unknown>,
+> = {
+  initial: NoInfer<State>;
```
alessbell, line 21: I'm curious about the decision to use `NoInfer` in the `MachineConfig` type. I say that because I find it easier to reason about when it's strongly typed since the keys and values are so similar, and it would also be a big maintainability win to have the type system catch typos where they're reversed like e.g. `devtoolsMachine.send({ type: "timedout" });` (should be `timeout`).

I see one downside re: `current.value` being inferred as `initial` though its value changes at runtime, so maybe there's a middle ground here, but curious about other challenges you maybe already ran into. A quick idea where the screaming case consts are strings:

```ts
const states = [
  INITIALIZED,
  CONNECTED,
  TIMED_OUT,
  NOT_FOUND,
  DISCONNECTED,
  RETRYING,
] as const;

const events = [CONNECT, TIMEOUT, CLIENT_NOT_FOUND, DISCONNECT, RETRY] as const;

type StatesUnionType = (typeof states)[number];
type EventsUnionType = (typeof events)[number];

interface DevToolsMachine<Context extends Record<string, unknown>> {
  initialContext?: Context;
  initial: typeof INITIALIZED;
  types: {
    events: { type?: EventsUnionType };
  };
  states: {
    [K in StatesUnionType]: {
      events?:
        | { [CONNECT]: typeof CONNECTED }
        | { [TIMEOUT]: typeof TIMED_OUT }
        | { [CLIENT_NOT_FOUND]: typeof NOT_FOUND }
        | { [DISCONNECT]: typeof DISCONNECTED }
        | { [RETRY]: typeof RETRYING };
    };
  };
}
```

src/application/components/ClientNotFoundModal.tsx
jerelmiller: Done in [8859408](https://github.com/apollographql/apollo-client-devtools/pull/1220/commits/8859408649cdcc678651630c45cfcd187e1f8b56). Please let me know if this captures the spirit of what is happening here 🙂

src/application/stateMachine.ts
jerelmiller: @alessbell the states can be inferred, but they are instead inferred from the rest of the object (the keys under `states`). If I allow `State` to be inferred from `initial`, it interferes with the type inference and it narrows the type of `State` to just the one declared value, rather than the union of all of them.

Here is a screenshot to show you the differences. Everything is type safe as you would expect otherwise (i.e. I can't declare an event in `events` that point to a state that doesn't exist 🙂)

**Without `NoInfer`**
Screenshot 2024-02-08 at 9 27 04 AM
Screenshot 2024-02-08 at 9 27 10 AM

**With `NoInfer<State>`**
Screenshot 2024-02-08 at 9 27 20 AM
Screenshot 2024-02-08 at 9 27 27 AM

Does that help?

jerelmiller: And to add to this on the event types themselves, these are also type safe. Here is an example where my event is "onnected" (I deleted the "c") and you can see that TypeScript complains in my machine config:
Screenshot 2024-02-08 at 9 33 18 AM

On the side of `machine.send(...)`, you also get the right event types listed here based on the `types` config:

Screenshot 2024-02-08 at 9 32 35 AM

alessbell: Ahh yep that makes sense! I see all the inference magic now :) We may want to add a comment about that since it's not immediately clear that the casting here

```
events: {} as
  | { type: "connect" }
  | { type: "timeout" }
  | { type: "disconnect" }
  | { type: "clientNotFound" }
  | { type: "retry" },
```

is responsible for typing the union of keys below.

alessbell (approved):
Excited to see this ship!

We just need a changeset - let's get this out at a new minor (or perhaps major?) so we can help users troubleshoot and narrow down any other remaining issues ASAP 🚀🎉

src/application/stateMachine.ts
jerelmiller: Ah ya I can do that. I totally stole this from [xstate](https://stately.ai/docs/machines#specifying-types) and thought it was a clever way for the types to be defined.

jerelmiller: Added in [6506ac8](https://github.com/apollographql/apollo-client-devtools/pull/1220/commits/6506ac89bb4a9c377249c308facaed1cff51502c)

PR #1225, opened 2024-02-14
Migrate to manifest v3 for Chrome

jerelmiller:
With the phase out of manifest v2 in [June 2024](https://developer.chrome.com/blog/resuming-the-transition-to-mv3), we need to migrate the Chrome extension to manifest v3. This PR updates the manifest file, introduces a service worker, and updates the build process to use v3 for chrome builds. Firefox remains on manifest v2 for now.


.github/workflows/release.yml
```diff
@@ -61,9 +62,9 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - name: Create zip files
+      - name: Create chrome zip file
         if: steps.changesets.outcome == 'success' && steps.changesets.outputs.published == 'true'
-        run: npm run dist chrome && npm run dist firefox
+        run: npm run dist:chrome
```
jerelmiller, line 67: These need to be separated out since each of them require a full rebuild, which wipes the `.zip` files.

jerelmiller:
From my testing, this seems to continue to work well in both browsers, so this should be ready for review 🙂

phryneas (approved):
Absolutely amazing that this is so little code! Great job!

Reviewed & tested both browsers with dev builds - everything is working for me.

webpack.config.mjs
```diff
@@ -49,20 +59,25 @@ export default (env) => {
         runLint: false,
         sourceDir: path.resolve(__dirname, "build"),
         startUrl: "http://localhost:3000",
-        target: env.TARGET,
+        target: WEB_EXT_TARGETS[target],
       })
     );
   }
 
+  const entry =
```
alessbell, line 67: Curious whether we've also tried shipping the same service_worker script at `background.scripts` for FF and using v3 for both?

package.json
```diff
@@ -14,16 +14,17 @@
...
     "start:server": "npm start --prefix development/server",
     "start:dev": "run-p start:client start:server",
-    "dist": "tsx build.ts",
+    "dist:chrome": "npm run build -- --env TARGET=chrome && tsx build.ts chrome",
+    "dist:firefox": "npm run build -- --env TARGET=firefox && tsx build.ts firefox",
     "build": "npm run clean && webpack --env NODE_ENV=production --progress",
     "build:dev": "npm run clean && webpack --env NODE_ENV=development --progress",
     "clean": "rimraf build && mkdir build && rimraf *.zip",
-    "changeset-publish": "npm run build && changeset publish",
+    "changeset-publish": "npm run build -- --env TARGET=firefox && changeset publish",
```
alessbell, line 22: Curious about `TARGET=firefox` here - I'm guessing it's just easier to ship just the FF artifacts to GitHub/npm? We may want to make the command name more specific to indicate that.

webpack.config.mjs
jerelmiller: I definitely went back and forth on this and I'm still a bit ensure whether I've made the right choice to split them. I was mostly thinking about this from the perspective that background scripts and service workers have [some differences](https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers) and if we needed to do something in one that isn't supported in the other, I didn't want to break one of the browsers. It's a future-proofing essentially.

That said, we don't have these differences now which means both the background script and service worker code is duplicated. This increases the maintenance burden when we _do_ need to make updates. In fact, I've started working on [updating](https://github.com/apollographql/apollo-client-devtools/pull/1226) the message passing between all the various pieces of the extension, and this problem is very apparent since I need to copy/paste all changes from one to the other.

My initial thinking with just moving Chrome to v3 and leaving firefox as-is was to ensure I didn't introduce a potential breakage in both browsers simultaneously. But, this PR ended up being less change than I initially thought, so perhaps we should just bite the bullet and update both.

(sorry for the rambling... just wanted to share my thought process here 😆)

package.json
jerelmiller: I added a check to ensure [`TARGET` is defined](https://github.com/apollographql/apollo-client-devtools/pull/1225/files#diff-504be09670485a92ff6bcb692041129e877eda9058a8c3114c42a53a0e0d1194R25-R27) as a means to make builds more intentional as to what we are building. This was also a bit of future-proofing. At some point I assume we'll want to support Edge and Safari and the `TARGET` makes it obvious what we are building.

Previous builds sort of already implicitly built it as the "firefox" target (though there really were no differences with the chrome build until now). My thinking here was that today it's build with manifest v2 which is now the firefox manifest, so this maintains compatibility with how its built today.

Thoughts? Would you prefer we default to a target if one isn't specified?

webpack.config.mjs
jerelmiller: Pushed an update that makes the code sharing a bit more robust: [663c402](https://github.com/apollographql/apollo-client-devtools/pull/1225/commits/663c4025a64638e662b842529086e087b04b935e). No need to copy/paste code, just import and use the existing background script!

I like this as a happy medium because I think the separate files highlights the differences in the two manifest/browser versions.

Thoughts on this?

jerelmiller: Firefox on manifest v3 is proving to be a bit trickier. The message passing seems to break. I think I'd like to look at this in a separate PR to figure out what's going on.

package.json
alessbell: > Previous builds sort of already implicitly built it as the "firefox" target

Previous builds weren't aware of any target afaik unless running in development mode since `WebExtPlugin` was the only one consuming `env.TARGET`, but maybe there's something I'm missing here?

jerelmiller: Ya that's correct. I probably used the term "firefox target" way too loosely 😂. `WebExtPlugin` was the only one that consumed it.

Do you know how the manifest file is used in the npm package outside of the extension? I'm actually a bit unsure how the npm package works on its own. Perhaps this is a good topic for our 1:1 today 🙂

alessbell: Sounds good :) Yeah, the manifest itself should only matter when shipping to an extension store.

There was a question about running it as a standalone package like Vue allows https://github.com/apollographql/apollo-client-devtools/discussions/1046. I think we'll likely have to make some changes to support that (the npm package has a few hundred DLs/week so perhaps it's already possible, I haven't checked), but let's handle that in a separate PR regardless.

.github/workflows/release.yml
alessbell: Makes sense. Thoughts on keeping the separation between build and dist commands so it's clear we're doing a fresh build here?

```suggestion
        run: npm run build -- --env TARGET=chrome && npm run dist -- chrome
```

alessbell (approved):
Thanks for tackling this long standing upgrade! Looking forward to getting our first v3 manifest out there 🎉🎉

PR #1226, opened 2024-02-15
Rework message passing

jerelmiller:
Given the number of recent reviews stating that the devtools doesn't work, I believe one of these issues is related to how we have message passing setup between the various pieces of the extension.

This PR makes the following adjustments to try and make this a bit more robust:
* Reworks the background script to map tabs and devtools scripts together in a single object.
* Moves away from the `Relay` class and toward a simpler solution that wraps a various message passing mechanism (window or port).
* Much stronger type safety between the messages sent in our extension.


.changeset/dirty-cars-work.md
```diff
@@ -0,0 +1,9 @@
+---
+"apollo-client-devtools": patch
```
jerelmiller, line 2: Minor? Not sure how best to classify this since its more or less a giant refactor. Perhaps the new message format classifies it as a minor?

src/Relay.ts
jerelmiller, line 1: 👋

src/application/components/Explorer/Explorer.tsx
```diff
@@ -64,31 +61,38 @@ function executeOperation({
...
 
-          window.addEventListener("message", checkForSubscriptionTermination);
-        } else {
-          observer.complete();
-        }
-      },
-      operationName,
-      !!isSubscription
-    );
+      if (payload.operationName !== operationName) {
```
phryneas, line 69: This seems prone to collisions - might it make sense to add a `uid` to the `explorerRequest` and `explorerResponse`, or even add a small `panelWindow.rpc` abstraction?

jerelmiller: Chatted in our 1:1. For transparency, here is what we discussed:

I absolutely agree with this and think this is a valid change. This PR is mostly focused on a 1:1 port of existing code over to `actor` from `Relay`, so we will do this in a followup PR.

PR #1245, opened 2024-03-06
Avoid serializing the payload before sending it in devtools messages

jerelmiller:
This removes the JSON parsing/serialization of the queries/cache/mutations payload sent in the devtools messages when communicating with various parts of the devtools. This is mostly redundant since `window.postMessage` uses the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) when sending data in message payloads.


phryneas (approved):
That's so much better, LGTM! :)

PR #1248, opened 2024-03-07
[WIP] Setup e2e tests with Playwright

jerelmiller:


jerelmiller:
The more I dig into this, this might actually be impossible to do in the way that I'm thinking. I'm primarily looking to test the message passing end-to-end, but every guide I come across tests either a popup or the extension changing something on the page.

You can either visit an app page, or you can visit an extensions html page, but using both together in an e2e test doesn't seem to be possible. I'm going to go ahead and close this as a failed experiment and will continue to think through how we can provide some automated testing for the message passing.

Issue #868, opened 2022-05-05
GCP Console crash


itsnotrisky:
**Intended outcome:**
Being able to open Google Cloud Console on chrome while the extension enabled.

**Actual outcome:**
The GCP Console produce error messages on console and prevent any page to be shown.

**How to reproduce the issue:**
Having the extension enabled, login to https://console.cloud.google.com/ and navigate to any menu. Recent support from reddit with further screenshots can be found [here](https://www.reddit.com/r/googlecloud/comments/uimswa/cant_open_any_menu_on_console/).

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser chrome
 - Browser version 100.0.4896.127
 - Extension version 4.0.0

jpvajda:
@itsnotrisky 👋 just to confirm, are you saying that until you disabled The Apollo Extension, you couldn't access the GCP console? I'm curious what happens if you try to re-enable it. Also I assume this was your Reddit comment (sponge bob consistency) 😄  so I'm just adding it for context as we are taking a look at this issue soon.


![Screen Shot 2022-05-17 at 12 26 44 PM](https://user-images.githubusercontent.com/1395158/168884778-0ec942cd-83fc-4a2f-a953-c7cef8c1f2fa.png)

itsnotrisky:
Correct, once I disabled the extension I can access GCP pages and as soon as I toggle it on from the extensions menu and refresh the same page on GCP Console it broke again. Similar note I have mentioned on the Reddit post is opening the GCP homepage/dashboard works fine but once I go to any sub menu i.e. to access Compute Engine that's when the error page shows up. 

And yeah I'm the OP there :D happy to add any details if you need to.

jpvajda:
@itsnotrisky Thanks for the additional details!

jeremiah-snee-openx:
This is happening again

jerelmiller:
@jeremiah-snee-openx thanks for bringing this to my attention! I'll dig in here soon to see if I can figure out which version broke the page.

phryneas:
This should be fixed in 4.8.2, which should land in the extension stores sometime soon.

PR #1249, opened 2024-03-07
Merge tabRelay code with tab

jerelmiller:
The `tabRelay` module exported a promise that resolved with the tab relay. It appears this behavior [was relied on](https://github.com/apollographql/apollo-client-devtools/commit/727f019c72a1694af6a99bfedbe90454ba1ef8c5#diff-5827f053ef6026472ecfad62402ce9534ae9ddab0a1ddf31bcccec71f2b169e3L12-L32) at one point in time, but removed in https://github.com/apollographql/apollo-client-devtools/commit/727f019c72a1694af6a99bfedbe90454ba1ef8c5. There is no reason this code needs to be wrapped in a promise now that nothing awaits it, so this PR removes the wrapping promise.


phryneas (approved):
Nice, one less file :)

jeremiah-snee-openx:
Not sure if this change is related but this issue seems to be happening again with the latest release. 
- https://github.com/apollographql/apollo-client-devtools/issues/868 

But this change moves code that was previously asynchronous and if failed would not have same impact on page being injected into

jerelmiller:
@jeremiah-snee-openx hmmm.... not sure this is related since the `Promise` callback is executed synchronously.

```js
console.log('before');
new Promise(async () => console.log('inside'));
console.log('after');
```

yields

```
before
inside
after
```

It looks like there was some [code](https://github.com/apollographql/apollo-client-devtools/commit/727f019c72a1694af6a99bfedbe90454ba1ef8c5#diff-5827f053ef6026472ecfad62402ce9534ae9ddab0a1ddf31bcccec71f2b169e3L12) removed in https://github.com/apollographql/apollo-client-devtools/commit/727f019c72a1694af6a99bfedbe90454ba1ef8c5 that waited for the promise to resolve before executing other stuff. Since that code was removed, it seemed a bit silly to maintain the promise callback since it was already synchronous.

---

That being said, I definitely can reproduce this. We switched injection mechanisms in [3.8.0](https://github.com/apollographql/apollo-client-devtools/releases/tag/v4.8.0). This seems more likely, but still baffles me since the manual code injection is now only there for [firefox](https://github.com/apollographql/apollo-client-devtools/blob/d006308948b521ebdb15c960c4af8fc2b37f5b66/src/extension/tab/tab.ts#L21-L41).

React devtools also [uses this technique](https://github.com/facebook/react/blob/7d6f1e3c13686272c1d76813992d4e782fcb84e6/packages/react-devtools-extensions/src/background/executeScript.js#L5-L36), but those seem to execute ok.

By chance, do you know what version of the devtools you upgraded from (no worries if not since the auto updating makes it a bit difficult to know 😅)? I'll dig in a bit and see if I can find the source of what broke this. Appreciate you bringing this to our attention!

jeremiah-snee-openx:
@jerelmiller your example looks synchronous because there is no delay inside the promise.

```js
const delay = ms => new Promise(res => setTimeout(res, ms));

console.log('before');
new Promise(async () => {
    await delay(200);
    console.log('inside')
});
console.log('after');
```

```
before
after
inside
```

jerelmiller:
@jeremiah-snee-openx apologies I probably should have been a bit more explicit in my comment about how it relates to the change in this PR. The `await` in `tabRelay` didn't come until the [very end](https://github.com/apollographql/apollo-client-devtools/pull/1249/files#diff-0b388e5604e2f1dc10903aaa268984d99fd5eeff8b6e81edfab89813d536f1d5L23) of the promise callback, so all of the setup code above it ran synchronously up to that point. Your example has an `await` before the `console.log('inside')` so it makes sense that this is logged last. Since there was nothing in `tab.ts` that `awaited` the promise exported from `tabRelay`, the promise really wasn't doing anything here. Sure, there was some async code executed inside the promise callback but there really wasn't any logic after the `await` that mattered to anything.

Regardless, https://github.com/apollographql/apollo-client-devtools/pull/1261 fixed the real issue with the GCP console crash which was released in [4.8.2](https://github.com/apollographql/apollo-client-devtools/releases/tag/v4.8.2). If you are still seeing issues with the latest versions, please open a new issue and we'd be happy to look 🙂

PR #1250, opened 2024-03-07
[wip] Client messaging test w/ Reactive vars

jerelmiller:
This works with changes in https://github.com/apollographql/apollo-client/pull/11647

This is meant to test the ability to send messages from the apollo client instance to the devtools. I will likely close this PR and discard these changes in the end, but for now this acts as a proof-of-concept for some of the changes happening in the client itself.


src/extension/tab/hook.ts
```diff
@@ -256,8 +262,12 @@ function initializeHook() {
     sendHookDataToDevTools("connectToDevtools");
   }
 
+  function handleClientMessage(message: ClientDevtoolsMessage) {
```
jerelmiller, line 265: This on its own seems to be working well, though per [my comment](https://github.com/apollographql/apollo-client/pull/11647/files#r1515609697) we would likely benefit from a message queue where we could replay messages here on initialization.

src/extension/tab/hook.ts
```diff
@@ -256,8 +262,12 @@ function initializeHook() {
...
+    tab.send(message);
```
jerelmiller, line 266: I'm very quickly realizing this "fire and forget" is too simplistic for something like Reactive vars. If the user never clicks into devtools, these messages get easily lost, which means the devtools scripts won't be able to pick up registered reactive vars.

Instead, we should do similar to queries/mutations and keep track of registered vars in a variable somewhere in this script. The devtools could then ask for any registered vars on initialization and sync that over.

jerelmiller:
Closing for now as this was just an experiment. This prompted the rpc work that landed in 4.9.0.

PR #1256, opened 2024-03-11
Add RPC helpers

jerelmiller:
Adds an RPC implementation to make it easier to ask for data between devtools extension types.

### DIFF
--- a/src/extension/__tests__/rpc.test.ts
+++ b/src/extension/__tests__/rpc.test.ts
@@ -48,6 +48,10 @@ function createTestAdapter(): TestAdapter {
     postMessage: jest.fn((message) => {
       proxy?.simulateMessage(message);
     }),
+    // Connects two adapters so that a postMessage from one adapter calls
+    // listeners on the proxy adapter. This isn't forwarding, but rather tries
+    // to simulate the window.postMessage(), window.addEventListener('message')
+    // behavior.
     connect: (adapter: TestAdapter) => {
       proxy = adapter;
     },
--- a/src/extension/__tests__/rpc.test.ts
+++ b/src/extension/__tests__/rpc.test.ts
@@ -1,11 +1,6 @@
 import { MessageAdapter } from "../messageAdapters";
 import { ApolloClientDevtoolsRPCMessage, MessageType } from "../messages";
-import {
-  RPCMessage,
-  createRPCBridge,
-  createRpcClient,
-  createRpcHandler,
-} from "../rpc";
+import { createRPCBridge, createRpcClient, createRpcHandler } from "../rpc";
 
 interface TestAdapter
   extends MessageAdapter<
@@ -67,7 +62,9 @@ function createBridge(adapter1: TestAdapter, adapter2: TestAdapter) {
 }
 
 test("can send and receive rpc messages", async () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
   // Since these are sent over separate instances in the real world, we want to
   // simulate that as best as we can with separate adapters
   const handlerAdapter = createTestAdapter();
@@ -85,7 +82,9 @@ test("can send and receive rpc messages", async () => {
 });
 
 test("resolves async handlers", async () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
   // Since these are sent over separate instances in the real world, we want to
   // simulate that as best as we can with separate adapters
   const handlerAdapter = createTestAdapter();
@@ -109,7 +108,9 @@ test("resolves async handlers", async () => {
 });
 
 test("does not mistakenly handle messages from different rpc calls", async () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
   const clientAdapter = createTestAdapter();
   const client = createRpcClient<Message>(clientAdapter);
 
@@ -132,7 +133,9 @@ test("does not mistakenly handle messages from different rpc calls", async () =>
 });
 
 test("rejects when handler throws error", async () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
   // Since these are sent over separate instances in the real world, we want to
   // simulate that as best as we can with separate adapters
   const handlerAdapter = createTestAdapter();
@@ -152,7 +155,9 @@ test("rejects when handler throws error", async () => {
 });
 
 test("rejects when async handler rejects", async () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
   // Since these are sent over separate instances in the real world, we want to
   // simulate that as best as we can with separate adapters
   const handlerAdapter = createTestAdapter();
@@ -170,9 +175,12 @@ test("rejects when async handler rejects", async () => {
 });
 
 test("can handle multiple rpc messages", async () => {
-  type Message =
-    | RPCMessage<"add", { x: number; y: number }, number>
-    | RPCMessage<"shout", { text: string }, string>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+    // while we're at it, let's have this one return a Promise in the definition
+    // it should not matter for the implementation
+    shout({ text }: { text: string }): Promise<string>;
+  };
 
   const handlerAdapter = createTestAdapter();
   const clientAdapter = createTestAdapter();
@@ -192,7 +200,9 @@ test("can handle multiple rpc messages", async () => {
 });
 
 test("only allows one handler per type", async () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
 
   const handle = createRpcHandler<Message>(createTestAdapter());
 
@@ -204,7 +214,9 @@ test("only allows one handler per type", async () => {
 });
 
 test("ignores messages that don't originate from devtools", () => {
-  type Message = RPCMessage<"add", { x: number; y: number }, number>;
+  type Message = {
+    add({ x, y }: { x: number; y: number }): number;
+  };
 
   const adapter = createTestAdapter();
   const handle = createRpcHandler<Message>(adapter);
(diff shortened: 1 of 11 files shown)


src/extension/rpc.ts
```diff
@@ -0,0 +1,191 @@
...
+type RPCRequestMessageFormat = {
+  type: string;
+  params: RPCParams;
+};
+
+type RPCResponseMessageFormat =
+  | { sourceId: number; result: unknown }
+  | { sourceId: number; error: unknown };
+
+export type RPCMessage<
```
jerelmiller, line 20: I don't know if there is a better way to do this. This is really just a container type to hold the values so that we can unwrap them in the various usages below (`RpcClient` and `createRPCHandler`)

src/extension/rpc.ts
```diff
@@ -0,0 +1,191 @@
...
+  __params: Params;
+  __returnType: ReturnType;
+};
+
+export interface RpcClient<
+  Messages extends RPCMessage<string, RPCParams, SafeAny>,
+> {
+  request: <TName extends Messages["__name"]>(
+    name: TName,
+    params: Extract<Messages, { __name: TName }>["__params"],
```
jerelmiller, line 35: I'm mostly happy with this, but this `params` is still a bit annoying. Right now this argumenet is required, even though its perfectly valid to send rpc calls without arguments (see the [`getClientOperations` call](https://github.com/apollographql/apollo-client-devtools/pull/1256/files#diff-4462d2ba22ebc0001138062ca5b0beef9eecbbce411838d1908bf9c4e5c2ec86R87) for example.) I'd like to make `params` only required if the params type has required fields. We will need to reorganize the function signature regardless since its awkward to set `options` with this function signature when `params` are optional.

src/extension/messages.ts
```diff
@@ -1,16 +1,41 @@
...
 
 export interface MessageFormat {
   type: string;
+  [key: string]: unknown;
 }
 
-export type ApolloClientDevtoolsMessage<Message extends MessageFormat> = {
+export const enum MessageType {
+  RPC = "rpc",
+  Event = "event",
```
jerelmiller, line 13: I'm not super in love with "event" as the name since this might get confusing with the browser built-in `Event` object. These are meant as fire-and-forget messages send through `actor`. [Worker RPC](https://github.com/DirtyHairy/worker-rpc) calls these "signals". Perhaps that makes more sense?

src/extension/rpc.ts
```diff
@@ -0,0 +1,191 @@
...
+    options?: { timeoutMs?: number }
+  ) => Promise<Extract<Messages, { __name: TName }>["__returnType"]>;
+}
+
+let nextMessageId = 0;
```
jerelmiller, line 40: I might consider adding something like [cuid2](https://github.com/paralleldrive/cuid2) to create the message ids. These are extremely subject to collisions and different areas of the devtools will initialize this value several times (i.e. its not shared). It _shouldn't_ be a problem for what we are using it for, but might be good to ensure `id` integrity regardless.

phryneas: We could also stick to browser apis to avoid pulling in another dependency:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID or, as I'm not sure if we're always in a "secure context", https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

src/extension/rpc.ts
```diff
@@ -0,0 +1,191 @@
...
+            source: "apollo-client-devtools",
+            type: MessageType.RPC,
+            id,
+            payload: { type: name, params },
+          });
+        }),
+      ]);
+    },
+  };
+}
```
phryneas, line 88: I would suggest that we try to avoid `Promise.race` here as it is very prone to [memory leaks](https://github.com/nodejs/node/issues/17469#issuecomment-685216777).


```suggestion
export function createRpcClient<
  Messages extends RPCMessage<string, RPCParams, SafeAny>,
>(
  adapter: MessageAdapter<
    ApolloClientDevtoolsRPCMessage<RPCRequestMessageFormat>
  >
): RpcClient<Messages> {
  return {
    request: (name, params, options) => {
      return new Promise((resolve, reject) => {
        const id = ++nextMessageId;

        const timeout = setTimeout(() => {
          removeListener();
          reject(new Error("Timeout waiting for message"));
        }, options?.timeoutMs ?? DEFAULT_TIMEOUT);

        const removeListener = adapter.addListener((message) => {
          if (
            isRPCMessage(message) &&
            "sourceId" in message.payload &&
            message.payload.sourceId === id
          ) {
            if ("error" in message.payload) {
              reject(message.payload.error);
            } else {
              resolve(message.payload.result);
            }

            removeListener();
            clearTimeout(timeout);
          }
        });

        adapter.postMessage({
          source: "apollo-client-devtools",
          type: MessageType.RPC,
          id,
          payload: { type: name, params },
        });
      });
    },
  };
}
```

phryneas: Also, writing it like this makes it easier to call `removeListener` in the case of a timeout.

phryneas (changes requested):
I love it!

Some comments and suggestions, but this is already looking great!

src/extension/rpc.ts
jerelmiller: Updated in [cc691ac](https://github.com/apollographql/apollo-client-devtools/pull/1256/commits/cc691ace76eaad32b2488d8eca9bd522b2d1da1a) by using `crypto.getRandomValues()`

src/extension/rpc.ts
jerelmiller: Updated in [cb7169f](https://github.com/apollographql/apollo-client-devtools/pull/1256/commits/cb7169fabdbca765650c8882e21d737415228357)

src/extension/devtools/devtools.ts
```diff
@@ -79,9 +83,17 @@ clientPort.send({ type: "connectToClient" });
 function startRequestInterval(ms = 500) {
   let id: NodeJS.Timeout;
 
+  async function getClientData() {
+    const payload = await rpcClient.request("getClientOperations", {});
+
+    if (panelWindow) {
+      panelWindow.send({ type: "update", payload });
+    }
+  }
```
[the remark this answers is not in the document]
jerelmiller: Done in [e66e02d](https://github.com/apollographql/apollo-client-devtools/pull/1256/commits/e66e02d14be86f1b7eb9c41de9a3501b7478d814)

src/extension/rpc.ts
jerelmiller: I ended up going with a bit different approach which avoids the awkwardness but also gives the more natural feel of allowing params to be a parameter array.

```ts
type Message = {
  add(x: number, y: number): number
}

const result = await client.withTimeout(1000).request('add', 1, 2);
```

Let me know what you think!

src/extension/rpc.ts
```diff
@@ -0,0 +1,182 @@
...
+      timeout = DEFAULT_TIMEOUT;
+
+      return new Promise<SafeAny>((resolve, reject) => {
+        const id = createId();
+
+        const timeout = setTimeout(() => {
+          removeListener();
+          reject(new Error("Timeout waiting for message"));
+        }, configuredTimeout);
+
```
phryneas, line 48: I'm not perfectly happy with the `timeout` variable, as this expects `request` to be called immediately after `withTimeout`.

What do you think about this?


```suggestion
export interface RpcClient<Messages extends MessageCollection> {
  readonly timeout?: number;
  withTimeout: (
    this: RpcClient<Messages>,
    timeoutMs: number
  ) => RpcClient<Messages>;
  request: <TName extends keyof Messages & string>(
    this: RpcClient<Messages>,
    name: TName,
    ...params: Parameters<Messages[TName]>
  ) => Promise<Awaited<ReturnType<Messages[TName]>>>;
}

const DEFAULT_TIMEOUT = 30_000;

export function createRpcClient<Messages extends MessageCollection>(
  adapter: MessageAdapter<RPCRequestMessage>
): RpcClient<Messages> {
  const client: RpcClient<Messages> = {
    timeout: DEFAULT_TIMEOUT,
    withTimeout(timeoutMs) {
      return { ...this, timeout: timeoutMs };
    },
    request(name, ...params) {
      return new Promise<SafeAny>((resolve, reject) => {
        const id = createId();

        const timeout = setTimeout(() => {
          removeListener();
          reject(new Error("Timeout waiting for message"));
        }, this.timeout);

```

src/extension/__tests__/rpc.test.ts
```diff
@@ -0,0 +1,565 @@
...
+  ).resolves.toEqual("first");
+
+  jest.advanceTimersByTime(29_000);
+
+  await expect(promise2).rejects.toEqual(
+    new Error("Timeout waiting for message")
+  );
+
+  jest.useRealTimers();
+});
```
phryneas, line 465: With my suggestions on `withTimeout`, this test would probably need to look like this.

```suggestion
test("resets timeout to default timeout after sending request", async () => {
  jest.useFakeTimers();
  type Message = {
    add(x: number, y: number): number;
  };

  const adapter = createTestAdapter();
  const client = createRpcClient<Message>(adapter);

  const t1 = client.withTimeout(1000);
  const t2 = client.withTimeout(2000);

  expect(client.timeout).toBe(30_000);
  expect(t1.timeout).toBe(1000);
  expect(t2.timeout).toBe(2000);

  const finished = new Set<Promise<number>>();

  const promise1 = t1.request("add", 1, 2);
  promise1
    .finally(() => finished.add(promise1))
    .catch(() => {
      /* prevent unhandled rejection warning */
    });

  const promise2 = t2.request("add", 1, 2);
  promise2
    .finally(() => finished.add(promise2))
    .catch(() => {
      /* prevent unhandled rejection warning */
    });

  jest.advanceTimersByTime(1000);

  // wait for a tick, just to be sure the `finally` block has time to run
  await Promise.resolve();
  expect(finished.has(promise1)).toBe(true);
  expect(finished.has(promise2)).toBe(false);

  await expect(promise1).rejects.toEqual(
    new Error("Timeout waiting for message")
  );

  jest.advanceTimersByTime(1000);
  // wait for a tick, just to be sure the `finally` block has time to run
  await Promise.resolve();
  expect(finished.has(promise2)).toBe(true);
  await expect(promise2).rejects.toEqual(
    new Error("Timeout waiting for message")
  );

  const promise3 = client.request("add", 1, 2);
  promise3
    .finally(() => finished.add(promise3))
    .catch(() => {
      /* prevent unhandled rejection warning */
    });

  jest.advanceTimersByTime(1000);

  // wait for a tick, just to be sure the `finally` block has time to run
  await Promise.resolve();
  expect(finished.has(promise3)).toBe(false);

  jest.advanceTimersByTime(29_000);

  // wait for a tick, just to be sure the `finally` block has time to run
  await Promise.resolve();
  expect(finished.has(promise3)).toBe(true);
  await expect(promise3).rejects.toEqual(
    new Error("Timeout waiting for message")
  );

  jest.useRealTimers();
});
```

src/extension/rpc.ts
jerelmiller: I like it. I noodled on this idea, but after seeing it in the test, I'm sold. Thanks!

jerelmiller: [2b0f59a](https://github.com/apollographql/apollo-client-devtools/pull/1256/commits/2b0f59a83a3672f27b880d73312a8e84f14ad28e)

PR #1257, opened 2024-03-11
use an object with functions as "RPC message" type

phryneas:
Triggered by https://github.com/apollographql/apollo-client-devtools/pull/1256#pullrequestreview-1926990431

### DIFF
--- a/src/extension/rpc.ts
+++ b/src/extension/rpc.ts
@@ -17,32 +17,20 @@ type RPCResponseMessageFormat =
   | { sourceId: number; result: unknown }
   | { sourceId: number; error: unknown };
 
-export type RPCMessage<
-  Name extends string,
-  Params extends RPCParams,
-  ReturnType,
-> = {
-  __name: Name;
-  __params: Params;
-  __returnType: ReturnType;
-};
+type MessageCollection = Record<string, (...parameters: [SafeAny]) => SafeAny>;
 
-export interface RpcClient<
-  Messages extends RPCMessage<string, RPCParams, SafeAny>,
-> {
-  request: <TName extends Messages["__name"]>(
+export interface RpcClient<Messages extends MessageCollection> {
+  request: <TName extends keyof Messages & string>(
     name: TName,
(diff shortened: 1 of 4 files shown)


src/extension/rpc.ts
```diff
@@ -17,46 +17,34 @@ type RPCResponseMessageFormat =
...
-export type RPCMessage<
-  Name extends string,
-  Params extends RPCParams,
-  ReturnType,
-> = {
-  __name: Name;
-  __params: Params;
-  __returnType: ReturnType;
-};
+type MessageCollection = Record<string, (...parameters: [SafeAny]) => SafeAny>;
```
phryneas, line 20: Using `RPCParams` here unfortunately causes all kind of downstream problems, especially with the "empty" case of `Record<string, never>` - so I've gone for a `SafeAny` here. Is there a specific reason this has to be an object and not just any serializable value>

jerelmiller: No reason it had to be an object. This was one of the final improvements I hadn't come back to :). I like this better.

src/extension/rpc.ts
```diff
@@ -17,46 +17,34 @@ type RPCResponseMessageFormat =
...
 
-export interface RpcClient<
-  Messages extends RPCMessage<string, RPCParams, SafeAny>,
-> {
-  request: <TName extends Messages["__name"]>(
+export interface RpcClient<Messages extends MessageCollection> {
+  request: <TName extends keyof Messages & string>(
```
jerelmiller, line 23: I'm curious the reason behind the `& string` here. Is that just to filter out any potential keys that aren't strings?

jerelmiller (approved):
I like this a lot. This feels so much more natural than the way I had set it up. Thanks so much for the improvement here! I'll keep iterating a bit to see if we can figure out how to make `params` a bit more natural, but otherwise this is amazing.

src/extension/rpc.ts
phryneas: Yeah, otherwise `keyof` will always be `| symbol`, which makes using this in other places very annoying.

PR #1259, opened 2024-03-12
Fix issue sending cache data to devtools when the cache contains a `URL` instance

### LINKED ISSUE #1258: unable to load devtool properly. Failed to execute 'postMessage' on 'Window'
Error occurs in dev mode when during the initial page load
```
Uncaught DOMException: Failed to execute 'postMessage' on 'Window': URL object could not be cloned.
    at Object.postMessage (chrome-extension://jdkknkkbebbapilgoeccciglkfbmbnfm/hook.js:1:12273)
    at Object.send (chrome-extension://jdkknkkbebbapilgoeccciglkfbmbnfm/hook.js:1:11729)
    at r (chrome-extension://jdkknkkbebbapilgoeccciglkfbmbnfm/hook.js:1:14709)
    at chrome-extension://jdkknkkbebbapilgoeccciglkfbmbnfm/hook.js:1:15291
```

the bug is likely caused by this PR - https://github.com/apollographql/apollo-client-devtools/pull/1245
**Intended outcome:**

**Actual outcome:**

No error

Start up any app in dev mode with Apollo Client and apollo-dev-tool enabled

- Browser [e.g. chrome, safari] chrome
- Browser version [e.g. 22]  122.0.6261.112 (arm64)
- Extension version [e.g. 3.0] 4.8.1

jerelmiller:
Fixes #1258

https://github.com/apollographql/apollo-client-devtools/pull/1245 removed the `JSON.stringify` call before sending the client data from the content script to the devtools script. While this mostly works well, there are some cases where this blows up, such as when the cache contains a `URL` instance. `URL` is a JSON serializable value, but unfortunately does not work with [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm). This PR reverts this change to serialize the cache before sending it to the devtools script.

### DIFF
--- a/src/extension/tab/hook.ts
+++ b/src/extension/tab/hook.ts
@@ -107,7 +107,8 @@ function initializeHook() {
         // references to `URL` instances which are not cloneable via
         // `structuredClone` (which `window.postMessage` uses to send messages).
         // `JSON.stringify` does however serialize `URL`s into strings properly,
-        // so this ensures that the cache data will be sent properly.
+        // so this should ensure that the cache data will be sent without
+        // errors.
         //
         // https://github.com/apollographql/apollo-client-devtools/issues/1258
         cache: JSON.stringify(hook.getCache()),
--- a/src/extension/tab/hook.ts
+++ b/src/extension/tab/hook.ts
@@ -102,7 +102,7 @@ function initializeHook() {
       payload: {
         queries: hook.getQueries(),
         mutations: hook.getMutations(),
-        cache: hook.getCache(),
+        cache: JSON.stringify(hook.getCache()),
       },
     });
   }
--- a/src/extension/tab/hook.ts
+++ b/src/extension/tab/hook.ts
@@ -102,6 +102,14 @@ function initializeHook() {
       payload: {
         queries: hook.getQueries(),
         mutations: hook.getMutations(),
+
+        // We need to JSON stringify the cache here in case the cache contains
+        // references to `URL` instances which are not cloneable via
+        // `structuredClone` (which `window.postMessage` uses to send messages).
+        // `JSON.stringify` does however serialize `URL`s into strings properly,
+        // so this ensures that the cache data will be sent properly.
+        //
+        // https://github.com/apollographql/apollo-client-devtools/issues/1258
         cache: JSON.stringify(hook.getCache()),
       },
     });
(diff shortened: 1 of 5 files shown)


phryneas (approved):
Bit sad we need to do this, but looks good.

src/extension/tab/hook.ts
```diff
@@ -97,7 +97,16 @@ function initializeHook() {
...
+
+        // We need to JSON stringify the cache here in case the cache contains
+        // references to `URL` instances which are not cloneable via
+        // `structuredClone` (which `window.postMessage` uses to send messages).
+        // `JSON.stringify` does however serialize `URL`s into strings properly,
+        // so this should ensure that the cache data will be sent without
+        // errors.
+        //
+        // https://github.com/apollographql/apollo-client-devtools/issues/1258
+        cache: JSON.stringify(hook.getCache()),
```
zhpadvoiski, line 109: doesn't seem that this fully fixed the issue. Since the payload includes queries, mutations, and a cache, and each of them could contain data that [doesn't work with structuredClone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone), shouldn't the whole thing be stringified? 
```ts
function sendHookDataToDevTools(eventName: "update" | "connectToDevtools") {
    tab.send({
      type: eventName,
      payload: JSON.stringify({
        queries: hook.getQueries(),
        mutations: hook.getMutations(),
        cache: hook.getCache(),
      }),
    });
  }
```

jerelmiller: Argh, you're right. I missed that `queries` and `mutations` have [`cachedData`](https://github.com/apollographql/apollo-client-devtools/blob/df5300f245b82e6efc15f616724874fbbb0b8a2d/src/extension/tab/helpers.ts#L14) on them. The `DocumentNode` and `variables` should always be fine otherwise. 

Thanks for pointing this out! I'll get another quick fix in place.

jerelmiller: This should be fixed with https://github.com/apollographql/apollo-client-devtools/pull/1268

jerelmiller: The above fix was released in 4.8.3 and should hit the web stores any time now. Thanks again for pointing this out!

PR #1264, opened 2024-03-12
add type-import lint rule

phryneas:
This should make it a bit more difficult for accidental bundle explosions like the one in #1262 to happen.


jerelmiller (approved):
💯

PR #1268, opened 2024-03-12
Fully fix issue sending data from client to devtools

jerelmiller:
Fully reverts changes from https://github.com/apollographql/apollo-client-devtools/pull/1245 that removed `JSON.stringify` from the entire client payload. Irregular cache values caused some issues when serializing on all fields.

### DIFF
--- a/src/extension/tab/hook.ts
+++ b/src/extension/tab/hook.ts
@@ -95,11 +95,11 @@ function initializeHook() {
     tab.send({
       type: eventName,
       // We need to JSON stringify the data here in case the cache contains
-      // references to `URL` instances which are not cloneable via
-      // `structuredClone` (which `window.postMessage` uses to send messages).
-      // `JSON.stringify` does however serialize `URL`s into strings properly,
-      // so this should ensure that the cache data will be sent without
-      // errors.
+      // references to irregular data such as `URL` instances which are not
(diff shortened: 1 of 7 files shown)


jerelmiller:
Noting here that @phryneas had a good idea to `JSON.parse(JSON.stringify(...))` in the content script before sending it. This should guarantee the value is serializable.

src/extension/tab/hook.ts
```diff
@@ -94,20 +94,21 @@ function initializeHook() {
...
-      },
+      // We need to JSON stringify the data here in case the cache contains
+      // references to irregular data such as `URL` instances which are not
+      // cloneable via `structuredClone` (which `window.postMessage` uses to
+      // send messages). `JSON.stringify` does however serialize `URL`s into
+      // strings properly, so this should ensure that the cache data will be
+      // sent without errors.
+      //
+      // https://github.com/apollographql/apollo-client-devtools/issues/1258
+      payload: JSON.parse(
```
jerelmiller, line 105: I'll note here that I think a potential future improvement here is to catch errors in case the cache can't be fully stringified. It would be a much nicer experience to show an error in the devtools panel, but we'll need to figure out a way to transport that error over. That is a bit out of scope for this PR.

PR #1271, opened 2024-03-14
feat: introduce simple apollo angular app

alessbell:
Brings over a demo Apollo + Angular starter app I built a while back which uses Apollo's HttpLink for `@defer` support.

Since its original purpose was to test `@defer`, I used the supergraph from https://github.com/apollographql/client-router-e2e-tests. I can bring it over as a standalone monograph which I can also add to this PR - thoughts?


jerelmiller (approved):
Thanks so much! This will be super helpful for smoke testing 🙂

PR #1278, opened 2024-03-19
Fix error serialization for rpc calls

jerelmiller:
While doing a bit of work to refactor another set of messages into an rpc call, I noticed that errors are not serialized properly in the rpc response. When doing the following:

```ts
handle('someRpcCall', () => Promise.reject(new Error('some error')))
```

This ended up as `{}` on the end that initiated the RPC call. Rather than trying to serialize the raw `Error` instance, this change serializes just the error message then rewraps it in an `Error` instance on the end listening for the response.


src/extension/rpc.ts
```diff
@@ -39,7 +40,7 @@ export function createRpcClient<Messages extends MessageCollection>(
 
         const timeout = setTimeout(() => {
           removeListener();
-          reject(new Error("Timeout waiting for message"));
+          reject(new Error(RPC_MESSAGE_TIMEOUT));
```
jerelmiller, line 43: I also added a new constant for this case to more easily compare when the timeout occurs.

We could also consider a `TimeoutError` class if we'd rather compare on the instance level rather than the message itself. For now I think this works well 🙂

src/extension/rpc.ts
```diff
@@ -119,7 +126,12 @@ export function createRpcHandler<Messages extends MessageCollection>(
           sourceId: id,
           result,
         });
-      } catch (error) {
+      } catch (e) {
+        const error =
```
jerelmiller, line 130: Noting here that @phryneas mentioned https://github.com/sindresorhus/serialize-error which could be much more useful since it maintains the stack trace and other properties.

jest.config.js
```diff
@@ -16,4 +16,8 @@ module.exports = {
       "<rootDir>/src/__mocks__/fileMock.js",
     "\\.(css|less)$": "<rootDir>/src/__mocks__/styleMock.js",
   },
+  transformIgnorePatterns: ["node_modules/(?!(serialize-error)/)"],
```
jerelmiller, line 19: `serialize-error` is an ESM package. This is needed to ensure Jest doesn't throw up on us.

phryneas (approved):
Love it!

phryneas (approved):

PR #1279, opened 2024-03-19
Convert call to connect to client to rpc call

jerelmiller:
We currently have 3 events that we use to handle the connect sequence to the client from devtools: `connectToClient`, `connectToClientTimeout`, and `clientNotFound`. With the introduction of the RPC message system, we are able to replace these with a single RPC call to connect to the client.

This requires an update to `findClient` in the hook to convert to an async function so that we can return the proper result from the RPC call. 

This PR depends on #1278 and is failing until that change is in place (which does not exist in this branch).


jerelmiller:
With the change to the state machine, I'm going to close this and rework it a bit from scratch. Doing this refactor work provided me with some good insight on some potential issues with the current setup, but I think starting with some changes to the state machine is a better place for this refactor.

PR #1280, opened 2024-03-19
Add `onLeave` callback for state machine transitions

jerelmiller:
Adds the ability to run a callback when leaving a particular state in a state machine. This can be done either by calling `machine.onLeave('state', callback)`, or by returning a function from `machine.onTransition(...)` that will be called when leaving the state.

The functionality in the machine is done, but I'd like to do a couple small refactorings after https://github.com/apollographql/apollo-client-devtools/pull/1279 is merged in.


jerelmiller:
Since we've moved to `@xstate/fsm`, there is no need for this anymore.

PR #1281, opened 2024-03-19
Inject error codes from Devtools

phryneas:
This would add all the basic functionality and could later be extended by downloading error codes for more versions from a CDN.

Open question: do we want to add this as an "opt-out" option to the Devtools or just have it always on>


package.json
```diff
@@ -3,6 +3,7 @@
   "description": "Developer tools for Apollo Client, with Studio Explorer integration, cache inspection, and query watching.",
   "license": "MIT",
   "version": "4.9.0",
+  "type": "module",
```
phryneas, line 6: As a result, this is now a module, and I had to generally change a few files around.

.github/workflows/update-errorcodes.yml
```diff
@@ -0,0 +1,39 @@
+name: Create PR for new error codes
+
+on: [workflow_dispatch]
```
phryneas, line 3: This would be triggered from the release workflows over in @apollo/client.

.github/workflows/update-errorcodes.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "18.x"
+      - run: |
+          cp package.json package.json.orig
+          npm run rebuild
+          if [ "$(wc -l < package.json.orig)" = "$(wc -l < package.json)" ]; then
```
phryneas, line 19: Checking if specifically lines were added by `npm run rebuild`. I've seen lines be reordered by `npm` in the past, so just looking for "changes" is not enough.

phryneas:
I had to make more changes as a result of the `type: "module"` and ended up porting the ESLint config over to the new config format.

phryneas:
On a positive sidenote, without all that downlevel transpilation, we save a lot of JS bundle size:
image

phryneas:
This seems to work in FF, but not in Chrome, so I'm moving it back to draft.

all-clients/errorcodes.json
```diff
@@ -0,0 +1,235 @@
...
+    "doc && doc.kind === \"Document\"",
+    "queryDef && queryDef.operation === \"query\"",
+    "doc.kind === \"Document\"",
+    "fragmentDef.kind === \"FragmentDefinition\"",
+    "mockedResponse.maxUsageCount > 0",
+    "!calledDuringRender()"
+  ],
+  "allFiles": [
+    null,
+    "@apollo/client/cache/inmemory/entityStore.js",
```
phryneas, line 152: Deliberately skipping the 0th index here so our byte arrays don't contain zeroes, which can help with "at a glance" detection of "obviously wrong" elements.

all-clients/restore-errorcodes.mjs
```diff
@@ -0,0 +1,29 @@
...
+  return result;
+}
+
+function decodeBase64(input) {
+  const binary = atob(input);
+  const bytes = new Uint16Array(binary.length);
+  for (let i = 0; i < binary.length; i++) {
+    bytes[i] = binary.charCodeAt(i);
+  }
+  return bytes;
```
phryneas, line 28: Doing it by hand because `Buffer` is not available in the browser, and this function is included in the extension *and* called from `node`.

phryneas:
I'll split this up into "modern output" and "error codes" PRs tomorrow and then it's really ready for review :)

.github/workflows/update-errorcodes.yml
phryneas: We're currently one release behind with this PR, which will give us a great opportunity to test this workflow :)

phryneas:
This is ready for isolated review now :)

.changeset/gentle-kiwis-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"apollo-client-devtools": minor
+---
+
+Add functionality to inject error and log messages.
```
jerelmiller, line 5:
```suggestion
Add the ability for devtools to provide full error and log messages without the need to need to call `loadDevMessages` or `loadErrorMessages`.
```

What about making this a bit more verbose to really make this feature stand out? For those reading the changelog, I think this might pique interest a bit more. Thoughts?

jerelmiller (approved):
This is SOOOOOOOOO dope. This is such a great use of devtools. Thanks so much for putting this together!

PR #1283, opened 2024-03-20
[Experiment] use `@xstate/fsm` instead of custom state machine implementation

phryneas:
This was triggered by #1280.

Maybe we could use an external state machine library instead of having to maintain our own?

This pulls in `@xstate/fsm`, which *should* be a similar size to our own implementation - let Relative-CI be the final judge to that.

As a nice bonus - here's the State Machine in the Stately Visualizer:

https://stately.ai/registry/editor/4c3732a9-9968-49f0-b377-22fa7864c799?machineId=ec297917-7ece-4052-8fac-652c31c7b768&mode=Design

Relevant context: `@xstate/fsm` is in a "deprecated-ish" state (quoting Andarist). 
It's very likely as safe to use as our own code, but I don't know if it'll get any updates in the future. I doubt it'll ever need any, though, and worst case there's always an upgrade path to XState (which I'm avoiding for bundle size here).


src/extension/devtools/devtools.ts
```diff
@@ -131,31 +134,30 @@ async function createDevtoolsPanel() {
...
-        devtoolsMachine.send({ type: "retry" });
+        devtoolsMachine.send("retry");
       });
 
-      devtoolsMachine.subscribe(({ state }) => {
-        panelWindow.send({ type: "devtoolsStateChanged", state: state.value });
+      const subscription = devtoolsMachine.subscribe(({ value }) => {
+        panelWindow.send({ type: "devtoolsStateChanged", state: value });
       });
+      unsubscribers.add(subscription.unsubscribe);
```
phryneas, line 150: This is the only real "change" of this PR - previously it seems that subscriptions were never unsubscribed, so I hope it's right to do that here?

phryneas:
This removes 2kb of our own code
image

And adds 4kb for `@xstate/fsm`
image

Worth the trade? 🤷

src/extension/devtools/devtools.ts
jerelmiller: I think this might actually introduce a bug because we only subscribe the first time the panel is created. Subsequent times will skip this code block, though there are several places that call `unsubscribeFromAll` (such as when the panel is hidden) so I think you might lose the subscription as soon as you switch to a different devtools panel.

The reason why I don't actually think we need this unsubscription is because closing the devtools window actually tears this stuff down, so there should be no possibility of memory leaks. As long as devtools is open, there is only ever one subscription here.

You can test this by adding a `console.log` above this, opening devtools, "Inspect element" on the Apollo panel, close devtools, then reopen. You'll notice that the window that inspects the Apollo panel disappears. If you open up devtools again and "Inspect element" on the apollo panel, you'll only ever see a single log for the subscription.

Thats the long way of saying, I'd just leave this as-is 🙂

jerelmiller (approved):
I think this is probably the right call given the number of years xstate has had developing a state machine library 🙂.

My only complaint about this change is that the nice callback actions on specific event transitions are a more difficult to achieve since its now replaced with a `switch` inside a single `subscribe` callback. But I think there is a way we could probably add an abstraction on top of this to get that style of callback back.

PR #1285, opened 2024-03-20
[Experiment follow-up] Add machine creation function to more easily handle actions

jerelmiller:
Followup to https://github.com/apollographql/apollo-client-devtools/pull/1283 that makes handling specific actions a bit nicer when the machine is split to the other file.

### DIFF
--- a/src/extension/devtools/devtools.ts
+++ b/src/extension/devtools/devtools.ts
@@ -142,10 +142,9 @@ async function createDevtoolsPanel() {
         devtoolsMachine.send("retry");
       });
 
-      const subscription = devtoolsMachine.subscribe(({ value }) => {
+      devtoolsMachine.subscribe(({ value }) => {
         panelWindow.send({ type: "devtoolsStateChanged", state: value });
       });
-      unsubscribers.add(subscription.unsubscribe);
 
       connectedToPanel = true;
     }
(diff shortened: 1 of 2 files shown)


src/extension/devtools/devtools.ts
```diff
@@ -144,10 +142,9 @@ async function createDevtoolsPanel() {
         devtoolsMachine.send("retry");
       });
 
-      const subscription = devtoolsMachine.subscribe(({ value }) => {
+      devtoolsMachine.subscribe(({ value }) => {
```
jerelmiller, line 145: This fixes the bug mentioned in https://github.com/apollographql/apollo-client-devtools/pull/1283#discussion_r1532946252

src/extension/devtools/devtools.ts
```diff
@@ -61,32 +85,6 @@ clientPort.on("clientNotFound", () => {
   devtoolsMachine.send("clientNotFound");
 });
 
-devtoolsMachine.subscribe(({ value }) => {
-  switch (value) {
```
jerelmiller, line 65: This was the main thing I was trying to get away from with this PR 🙂. `switch` has its place, but this could get unwieldy as we add more states (in fact I have another state or two in mind already).

jerelmiller:
Spoke with @phryneas offline and we are good with this change. Will merge this and the other PR together 🙂

PR #1286, opened 2024-03-21
Build for a more modern target

phryneas:
image
🎉


package.json
```diff
@@ -27,7 +28,7 @@
     "firefox": "npm run build:dev -- --watch --env TARGET=firefox",
     "format": "prettier --write .",
     "check:format": "prettier --check .",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "lint": "ESLINT_USE_FLAT_CONFIG=true yarn eslint src",
```
jerelmiller, line 31: TIL

jerelmiller (approved):
Do you mind adding a changeset before merging this? Otherwise looks great!

tsconfig.json
```diff
@@ -1,13 +1,15 @@
 {
   "compilerOptions": {
     "allowJs": true,
+    "checkJs": true,
```
jerelmiller, line 4: I think this might be causing problems with the `all-clients` package since it uses `.mjs` with import attributes. I can get this to build if I remove this line (though I understand why you enabled it). Just an FYI

phryneas: Yeah... let's skip it for now.

PR #1287, opened 2024-03-21
don't use `gql` in `hook.js`

phryneas:
This moves the usage of `gql` out of the injected `hook.js` into the DevTools panel.

That should significantly reduce the amount of injected code.

image
🎉


src/application/components/Explorer/postMessageHelpers.ts
```diff
@@ -84,7 +83,7 @@ export type ExplorerRequest = MessageEvent<{
   name: "ExplorerRequest";
   operationId: string;
   operation: string;
-  variables: JSONValue;
```
phryneas, line 87: This is definitely never an array or a primitive value ;)

src/application/components/Explorer/Explorer.tsx
```diff
@@ -43,20 +44,21 @@ function executeOperation({
...
-      variables,
-      fetchPolicy,
+    panelWindow.send({
+      type: "explorerRequest",
+      payload: {
+        operation: gql(operation),
+        operationName,
+        variables,
+        fetchPolicy,
+      },
```
phryneas, line 59: Actually, no need to `stringify` this one.

Everything here already came in from a `postMessage`, so it's transportable. The only possible exception would be the `DocumentNode` itself, but that seems to work nicely with `structuredClone`.

src/application/components/Explorer/Explorer.tsx
```diff
@@ -220,7 +222,7 @@ export const Explorer = ({
           const observer = executeOperation({
             operation: event.data.operation,
             operationName: event.data.operationName,
-            variables: event.data.variables,
+            variables: event.data.variables ?? undefined,
```
phryneas, line 225: replacing a potential `null` with `undefined`.

jerelmiller (approved):
Mind adding a changeset for this one? Otherwise this looks great! What a great size savings here. Simple change but super effective 🚀

PR #1288, opened 2024-03-21
[WIP] add a memory panel

phryneas:


src/application/index.tsx
```diff
@@ -73,9 +75,18 @@ const cache = new InMemoryCache({
...
 const cacheVar = makeVar<string | null>(null);
 export const client = new ApolloClient({
   cache,
+  resolvers: {
+    Query: {
+      memoryInternals(): Promise<MemoryInternals | undefined> {
+        return rpc.request("getMemoryInternals");
+      },
+    },
+  },
```
phryneas, line 89: I don't know if this is genius or incredibly stupid, but it might be very interesting in combination with `pollInterval`.
@jerelmiller do you have an opinion on this?

src/extension/devtools/devtools.ts
```diff
@@ -131,6 +134,13 @@ async function createDevtoolsPanel() {
   panel.onShown.addListener((window) => {
     panelWindow = getPanelActor(window);
 
+    unsubscribers.add(
+      createRPCBridge(
+        createPortMessageAdapter(port),
+        createWindowMessageAdapter(window)
+      )
+    );
+
```
phryneas, line 143: Not 100% sure if this is the right place to create this bridge 😬

src/application/index.tsx
jerelmiller: I think this is totally fine! And agreed. The more I'm working with devtools, the more I think we can move some of the code in `devtools.ts` over to the React application, This falls in line with my thinking there so I dig it.

src/extension/devtools/devtools.ts
jerelmiller: I would make two changes here:

1. Move the `createRPCBridge` call inside the the `if (!connectedToPanel)` below.
2. Remove the `unsubscribers.add` call around this

The reasoning here is that we still want to allow messages to travel between the background port and the panel window even when the panel is hidden (e.g. the user navigated away by clicked the "Elements" tab). I can think of some cases like opening the Apollo tab, then quickly switching away from it, but still allowing the injected script to report back whether it found a client or not (which will eventually be an RPC call as well). If we unsubscribe the bridge, any message that gets sent while the panel is hidden will be dropped.

The `if (!connectedToPanel)` should only ever execute the first time the panel is created, so we shouldn't have to worry about this script accidentally creating too many bridges.

jerelmiller:
Closing as this will be superceded by https://github.com/apollographql/apollo-client-devtools/compare/main...jerel/memory-internals

PR #1313, opened 2024-04-03
fixup errorcodes workflow

phryneas:
`.md` is not a good filename for a changeset 😅


.github/workflows/update-errorcodes.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             echo "No new error codes were added."
             exit 0
           fi
-          cat <<EOF > ../.changeset/`md5 < errorcodes.json`.md
+          cat <<EOF > ../.changeset/errorcodes.md
```
phryneas, line 24: I believe it should be fine to always use the same name here, right @alessbell?

alessbell: Generally yes if we can assume there would only ever be one PR with an errorcodes.md open at a time, but if you ever introduce a prerelease workflow I'd change this so the file name is unique.

phryneas: Then let's revisit this if we ever get to the point of doing prereleases with the devtools ^^

phryneas: As for the "ever only one" part - since they will all have the same message, it's probably even better that we don't repeat that over and over in the case of multiple PRs - it's not really important for the release message to know if error messages were updated once or three times between releases ^^

PR #1335, opened 2024-04-20
Fix memory leak issue with dev tools

### LINKED ISSUE #1332: Serious memory leak issue with the devtools extension
**Intended outcome:**
the extension should run stable and not crash.

**Actual outcome:**
The extension would crash several minutes after opening it. 

I work on a project with pretty big schema, and the crash would happen in several minutes.
I think with a smaller schema, this can still be verified by opening dev console for the extension, go to memory tab,
check memory  consumption of the devTools.html and panel.html
the value would increase with time. 
on my project it would reach 3 GB pretty quickly and crash. 

- OS: Mac OS 14.4
- Browser: Chrome 
- Browser version 123.0.6
- Extension version 4.11.3

hehaijin:
This is to fix the memory leak identified in #1332

Each copy of client data to extension is wrapped inside a promise, 
and this data is supposed to be removed and GCed but right now it's not. 

in src/extension/messageAdapters.ts `createPortMessageAdapter` 
the line `port.onDisconnect.addListener` creates 
a closure that holds the reference to `handler` passed in, and only release when disconnect, 
which may not happen very soon. 
And in src/extension/rpc.ts `createRpcClient` function, the the handler holds 
a reference to the promise that get data from client side.
so the promise data never got garbage collected.

Removed this event handler from onDisconnect to fix the memory leak issue.

May need to address how to properly remove it in `onDisconnect`, but it's lower priority. 
closes #1332

### DIFF
--- a/src/extension/messageAdapters.ts
+++ b/src/extension/messageAdapters.ts
@@ -18,9 +18,6 @@ export function createPortMessageAdapter<
   return {
     addListener(listener) {
       port.onMessage.addListener(listener);
-      port.onDisconnect.addListener(() => {
-        port.onMessage.removeListener(listener);
-      });
 
       return () => {
         port.onMessage.removeListener(listener);
--- a/.changeset/rich-balloons-prove.md
+++ b/.changeset/rich-balloons-prove.md
@@ -1,0 +1,5 @@
+---
+"apollo-client-devtools": patch
+---
+
+Fixed a memory leak issue with devtools


src/extension/messageAdapters.ts
```diff
@@ -18,9 +18,6 @@ export function createPortMessageAdapter<
   return {
     addListener(listener) {
       port.onMessage.addListener(listener);
-      port.onDisconnect.addListener(() => {
```
hehaijin, line 21: event handler would hold listener and prevent it from garbage collected.

jerelmiller:
Hey @hehaijin 👋 

Thanks so much for this PR! I don't actually remember why I added the code to remove the `onMessage` event listener in `onDisconnect`. Googling around, I can't seem to find anything that specifies that this is needed. Perhaps I saw it somewhere and blindly copied it 😅.

In any case, this PR looks good and I think its a small enough change that should be fine. If you wouldn't mind adding a changeset to this PR, I'd be happy to get this merged and released. Appreciate all the digging you've been doing into the memory consumption of the devtools! We'd welcome any improvements you find!

hehaijin:
Thanks @jerelmiller  for reviewing. Added changeset.

hehaijin:
@jerelmiller  I have some thoughts regarding performance posted in discussion: https://github.com/apollographql/apollo-client-devtools/discussions/1337
it would be nice to know what your guys think.

jerelmiller (approved):
🎉 Thanks again so much for this PR!

PR #1352, opened 2024-05-06
Schedule next request to get data after previous request finishes

jerelmiller:
Small optimization to only trigger a new fetch of data after the previous completes. In case we are dealing with large payloads that take a significant amount of time to copy over, this prevents us from potentially overlapping the next request before the previous finishes.


src/extension/devtools/devtools.ts
```diff
@@ -98,14 +98,15 @@ function startRequestInterval(ms = 500) {
         payload: await rpcClient.request("getClientOperations"),
       });
     }
+
+    id = setTimeout(() => getClientData(), ms);
```
phryneas, line 102:
```suggestion
    id = setTimeout(getClientData, ms);
```
I think we don't need the function definition here

phryneas: Also, should we wrap the whole block in try..finally to ensure an exception doesn't break the loop?

jerelmiller: Thats a good idea!

PR #1353, opened 2024-05-06
Reduce bundle size of error codes by creating lookup of errors

jerelmiller:
There are many versions that share the same encoded error contents. Rather than creating each of these per-version, this change creates a lookup by id and modifies the version to point to the id of the errors it represents. This means we only need to save new hashes when errors are modified from the client.


phryneas (approved):
Looks good to me :)

PR #1362, opened 2024-05-09
Remove schema data from client message to devtool

hehaijin:
Schema data is requested by 'IntrospectionQuery'
for showing documentation in dev tool for the explorer tab.

It is removed from queries and cache tab as no need to show there.
However, for each message to sync client data to dev tools,
the schema data is still included in each message, and needs to be
stringified/parsed/processed/GCed for each message.

This creates big performance issue when schema is big.
by removing it I see a big reduction in memory and cpu usage.


hehaijin:
in my case rough estimate in this test is that all other queries/mutations are around 100kb
while our schema is > 10MB.
So the effect is very obvious.
before
image
after
image

hehaijin:
@jerelmiller  could you take a look?

jerelmiller:
Hey @hehaijin this is a really great find! The reasoning here makes sense. The only thing that uses that introspection query is the explorer since it uses the apollo client instance to send queries. No reason we need to hang onto that.

Would you mind adding a changeset? I'll be happy to merge and get this released as soon as you do 🙂. Thanks so much for looking into this! I'm sure this will be super helpful for you app.

hehaijin:
@jerelmiller  Thanks for review. Added change set.

jerelmiller (approved):
🔥 Thanks again! Looking forward to getting this one out.

PR #1375, opened 2024-05-17
Modify legacy registration mechanism

jerelmiller:
This is a precursor to some work I'm doing to make client discovery a bit less chaotic. Uses `Object.defineProperty` for the global property to monitor its setter and register the client with the devtools.

### DIFF
--- a/src/extension/tab/hook.ts
+++ b/src/extension/tab/hook.ts
@@ -308,8 +308,11 @@ Object.defineProperty(window, "__APOLLO_CLIENT__", {
   get() {
     return globalClient;
   },
-  set(client: ApolloClient<SafeAny>) {
-    registerClient(client);
+  set(client: ApolloClient<SafeAny> | undefined) {
+    if (client) {
+      registerClient(client);
+    }
+
     globalClient = client;
   },
 });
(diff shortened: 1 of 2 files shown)


src/extension/tab/hook.ts
```diff
@@ -301,4 +301,19 @@ if (Array.isArray(preExisting)) {
   (preExisting as Array<ApolloClient<any>>).forEach(registerClient);
 }
 
-findClient();
```
jerelmiller, line 304: We no longer need this on initialization since we register the client immediately if is available before the script runs and handles when the value is assigned from a legacy client.

src/extension/tab/hook.ts
```diff
@@ -91,12 +91,6 @@ window.onbeforeunload = () => {
   tab.send({ type: "disconnectFromDevtools" });
 };
 
-window.addEventListener("load", () => {
-  if (hook.ApolloClient) {
```
jerelmiller, line 95: This was almost never true, but with the changes in this PR, we don't need it anyways since we will register the client immediately if its available when this content script loads, or register it when the client is assigned on the window.

phryneas (approved):

PR #1378, opened 2024-05-19
Fix intermittent issue where a page refresh or navigation might mistakenly connect but show no data

jerelmiller:
Our current disconnect event sent to the devtools prompts devtools to try and connect to the client again by sending the `connectToClient` message. This caused some weird issues where sometimes that event was triggered so fast that it re-discovered the old client before the page was fully unloaded. At times this seemed to cause issues where it was reported that the client was connected, but you saw no data.

This PR introduces a new `registerClient` event sent by the content script that is fired when a client registered with the devtools to distinctly identify the event as such. Now when a disconnect happens, rather than pinging the content script over and over to try and determine if a client is connected, it will instead start a new timer, wait 10 seconds, and if the `registerClient` event hasn't occured within that time, will set itself to the "not found" state. This makes this behavior much more reliable between page refreshes and navigation.

This PR also makes a small improvement by detecting if the client is able to connect again after it has found itself in the "not found" state. If, for example, you navigate to a page without Apollo loaded, then navigate back using the back button, it will re-connect itself and show the data in the devtools without first having to click the "retry" button.

If you want to validate some cases yourself, try doing these few things:
* Reloading the page
* Navigate to url without Apollo (e.g. github.com)
  * Ensure it shows "disconnected", then you get "not found" after ~10 sec
  * Hit the back button, ensure it reconnects to the app that has an Apollo Client instance
  * Hit forward, should see "disconnected"
  * Hit back before the "not found" message is shown. You should see it connect again.

### DIFF
--- a/src/application/machines.ts
+++ b/src/application/machines.ts
@@ -87,7 +87,7 @@ export function createDevtoolsMachine({ actions }: { actions: Actions }) {
             timeout: "timedout",
             clientNotFound: "notFound",
           },
-          entry: ["unsubscribeFromAll", "connectToClient"],
+          entry: ["unsubscribeFromAll"],
         },
         timedout: {},
         notFound: {
--- a/src/application/machines.ts
+++ b/src/application/machines.ts
@@ -93,6 +93,7 @@ export function createDevtoolsMachine({ actions }: { actions: Actions }) {
         notFound: {
           on: {
             retry: "retrying",
+            connect: "connected",
           },
           entry: "unsubscribeFromAll",
         },
--- a/src/extension/messages.ts
+++ b/src/extension/messages.ts
@@ -82,6 +82,7 @@ type ExplorerSubscriptionTerminationMessage = {
 };
 
 export type ClientMessage =
+  | { type: "registerClient" }
   | { type: "clientNotFound" }
   | { type: "connectToClient" }
   | { type: "connectToClientTimeout" }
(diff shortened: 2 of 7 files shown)


src/application/machines.ts
```diff
@@ -87,12 +87,13 @@ export function createDevtoolsMachine({ actions }: { actions: Actions }) {
             timeout: "timedout",
             clientNotFound: "notFound",
           },
-          entry: ["unsubscribeFromAll", "connectToClient"],
+          entry: ["unsubscribeFromAll"],
```
jerelmiller, line 90: This ended up working very strangely because sometimes the `connectToClient` event was sent before a page was fully unloaded, which caused some false-positives on determining when the client reconnected. Instead, we now just rely on the content script sending the `registerContent` event to determine when to navigate back to the `connected` state

src/extension/messages.ts
```diff
@@ -82,9 +82,9 @@ type ExplorerSubscriptionTerminationMessage = {
 };
 
 export type ClientMessage =
+  | { type: "registerClient" }
   | { type: "clientNotFound" }
   | { type: "connectToClient" }
-  | { type: "connectToClientTimeout" }
```
jerelmiller, line 87: This wasn't used so no reason to keep it around.

src/application/machines.ts
phryneas:
```suggestion
          entry: "unsubscribeFromAll",
```

phryneas (approved):
Read the code, reproduced the bug, reproduced that this fixes it - LGTM :)

PR #1379, opened 2024-05-20
Push events from the client instead of polling for them

jerelmiller:
Reintroduces the `__actionHookForDevtools` hook which allows us to push updates from the client to the devtools. This removes the need for polling from the devtools for updates continuously.

As a result, the polling code is (mostly) gone except for the handling of one edge case where queries kicked off right after the client is registered are not reported via the action hook. We will poll for a brief time until we know the client starts pushing data again or it disconnects.

This change has the benefit that changes from the client are reflected immediately in the devtools UI rather than waiting for the next tick of the poll interval so it feels a bit snappier.

To test this, try a few different situations:
* Navigating around the app to ensure the listed queries change
* Reload the page to ensure the data is shown in the devtools after refresh
* Destroy/recreate the client


phryneas (approved):
Read the code and tested it - this is amazing! :)

jerelmiller:
Going to push out the other two changes, then will get to this next week. Spacing them will ensure the other release doesn't break anything.

PR #1389, opened 2024-05-29
Remove unneeded setup/teardown of message forwarding in the panel

jerelmiller:
Small optimization to the setup of the panel. We only need to create the port forwarding once rather than setting it up then tearing it down anytime the user navigates away from the devtools tab. Since we no longer poll for changes as of #1379 this was even more unnecessary.

### DIFF
--- a/src/extension/devtools/devtools.ts
+++ b/src/extension/devtools/devtools.ts
@@ -122,10 +122,6 @@ async function createDevtoolsPanel() {
     "panel.html"
   );
 
-  let removeExplorerForward: () => void;
-  let removeSubscriptionTerminationListener: () => void;
-  let removeExplorerListener: () => void;
-
   panel.onShown.addListener((window) => {
     panelWindow = getPanelActor(window);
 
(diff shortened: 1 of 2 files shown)


src/extension/devtools/devtools.ts
```diff
@@ -161,26 +157,17 @@ async function createDevtoolsPanel() {
...
+
       connectedToPanel = true;
     }
 
     if (devtoolsMachine.state.value === "initialized") {
       clientPort.send({ type: "connectToClient" });
       startConnectTimeout();
     }
-
-    removeExplorerForward = clientPort.forward("explorerResponse", panelWindow);
```
jerelmiller, line 172: This was run every time the user changed tabs in the devtools, then navigated back to Apollo devtools. Since this just sets up forwarding between the two "actors" so that the message passing can communicate between the two, no reason why we needed to cut off these messages when the user navigates away. Now we'll just set them up the first time the panel is created and call it good.

alessbell (approved):
Nice optimization 💯

PR #1390, opened 2024-05-29
Automatically reconnect the ports when they disconnect

jerelmiller:
When a port disconnects, we will now try to automatically reconnect it to the background script. This should help in situations where the browser might disconnect the service worker [due to inactivity](https://developer.chrome.com/docs/extensions/whats-new#m110-sw-idle).

### DIFF
--- a/.changeset/late-crabs-rule.md
+++ b/.changeset/late-crabs-rule.md
@@ -1,0 +1,5 @@
+---
+"apollo-client-devtools": patch
+---
+
+Automatically reconnect the ports when they disconnect.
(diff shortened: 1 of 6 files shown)


alessbell (approved):
Makes sense, especially given the 30 second idle timeout introduced in chrome 110 - hadn't seen that before. 🚀

jerelmiller:
@alessbell exactly! I stumbled upon it and it made a lot of sense why users might feel like it stops working after a time. Thanks for the review!

PR #1391, opened 2024-05-29
Use `browser.devtools.network.onNavigated` to detect page navigation

jerelmiller:
This moves the detection of page navigation from the injected script to the devtools by utilizing the `browser.devtools.network.onNavigated` API.


src/extension/messages.ts
```diff
@@ -90,7 +90,7 @@ export type ClientMessage =
       type: "connectToDevtools";
       payload: ClientContext;
     }
-  | { type: "disconnectFromDevtools" }
+  | { type: "clientTerminated" }
```
jerelmiller, line 93: With the change in this PR, it felt appropriate to rename this message since a terminated client is the only thing that sent this message. This now better represents the actual thing that happened.

src/extension/devtools/devtools.ts
```diff
@@ -185,3 +187,5 @@ async function createDevtoolsPanel() {
 }
 
 createDevtoolsPanel();
+
+browser.devtools.network.onNavigated.addListener(disconnectFromDevtools);
```
alessbell, line 191: Do we need to call `removeListener` manually, or will this get removed automatically on e.g. hard page refresh?

alessbell (approved):

src/extension/devtools/devtools.ts
jerelmiller: We don't need to worry about cleaning this up since closing devtools should terminate this script. 

This handler registers when devtools is first opened and fires when a user either refreshes the page or navigates to a new page (non `pushState` navigation that is). I just tested this by adding a `console.log`, closing the devtools, reopening and I only saw it executed once (in other words, the listener got cleaned up for me).

The only case where we'd need to think about doing some kind of cleanup on listeners is in the content scripts if for example the chrome extension updates in the background and the service worker is terminated and recreated (something I don't think we handle very well right now). Anyways, all this to say, we should be fine to register this once and don't need to worry about cleaning this up.

alessbell: Makes sense!

PR #1392, opened 2024-05-29
Clean up some redundant logic for connecting to the client

jerelmiller:
With the updated registration mechanism added in #1375, we can simplify the way the connect timeouts are handled and remove some redundant logic for trying to find the client. This PR simplifies this mechanism to remove some redundancy.


src/extension/tab/hook.ts
```diff
@@ -223,28 +214,6 @@ tab.on("explorerRequest", (message) => {
   }
 });
 
-/**
- * Attempt to find the client on a 1-second interval for 10 seconds max
- */
-let interval: NodeJS.Timeout;
-function findClient() {
```
jerelmiller, line 230: We no longer need this at all since we send a `registerClient` message any time the client is registered which will connect it for us.

src/extension/tab/hook.ts
```diff
@@ -117,18 +117,9 @@ function getClientData() {
...
-    type: eventName,
-    payload: getClientData(),
-  });
-}
-
 tab.on("connectToClient", () => {
   if (hook.ApolloClient) {
-    sendHookDataToDevTools("connectToDevtools");
-  } else {
-    findClient();
```
jerelmiller, line 131: The devtools script will now try to connect and then wait up to 10 seconds for a `registerClient` or `connectToDevtools` message before giving up and showing the not found modal. Since we can rely on those messages instead, we don't need to try and find a client when this `connectToClient` message is sent.

In the future, this `connectToClient` could likely be replaced with an RPC call to simplify this a bit further. This adds a bit of complexity though to do this in this PR so I'll do this separately to keep this PR clean.

src/extension/devtools/devtools.ts
```diff
@@ -164,11 +144,6 @@ async function createDevtoolsPanel() {
       connectedToPanel = true;
     }
 
-    if (devtoolsMachine.state.value === "initialized") {
```
jerelmiller, line 167: This is no longer needed since the rest of the messages are now robust enough to not have to "retry" connecting when the panel is first shown 🎉

alessbell (approved):
Nice that we no longer need `findClient`! 🚀

PR #1394, opened 2024-05-30
Add history navigation to cache tab

jerelmiller:
Adds a forward/back button with history to the cache tab. This should make it easier to navigate back and forth when digging into more complex objects that have lots of nested normalized data. 

Previously getting back to where you were would require you to start back at either the root query or have to manually search the cache for the entry you came from.

Screenshot 2024-05-29 at 8 51 59 PM
Screenshot 2024-05-29 at 8 55 53 PM
Screenshot 2024-05-29 at 8 56 54 PM


src/application/components/Alert.tsx
```diff
@@ -0,0 +1,46 @@
...
+  children?: ReactNode;
+  className?: string;
+}
+
+const alert = cva(
+  ["px-4", "py-3", "border-l-4", "rounded", "flex", "gap-2", "items-start"],
+  {
+    variants: {
+      variant: {
+        error: [
```
jerelmiller, line 20: For now we only need the `error` variant, so I've only implemented this one. We can add others as we need them.

src/application/components/Cache/Cache.tsx
```diff
@@ -39,9 +46,11 @@ function filterCache(cache: Cache, searchTerm: string) {
   );
 }
 
+const history = new History("ROOT_QUERY");
```
jerelmiller, line 49: Instantiating this outside of the component allows us to keep the history when changing tabs, otherwise this resets anytime the `Cache` component unmounts.

src/application/components/Cache/Cache.tsx
```diff
@@ -70,20 +84,61 @@ export function Cache() {
               value={searchTerm}
             />
             <div className="overflow-auto h-full">
-              <EntityList
-                data={filteredCache}
-                selectedCacheId={cacheId}
-                setCacheId={setCacheId}
-                searchTerm={searchTerm}
-              />
+              <List>
```
jerelmiller, line 87: This component was fairly small and "EntityList" wasn't a super great name for this anyways, so I went ahead and inlined the implementation here and removed the old component.

src/application/components/Tooltip/Tooltip.tsx
```diff
@@ -5,12 +5,18 @@ interface TooltipProps {
...
-export function Tooltip({ content, children, side = "bottom" }: TooltipProps) {
+export function Tooltip({
+  content,
+  children,
+  delayDuration,
+  side = "bottom",
+}: TooltipProps) {
   return (
-    <Root>
+    <Root delayDuration={delayDuration}>
```
alessbell, line 19: Nice UX detail 💯

alessbell (approved):
Great feature that will make navigating the cache much nicer! 🚀

PR #1409, opened 2024-06-07
Minor refactor to remove some more redundancy

jerelmiller:
This does another small refactor to remove some redundancy on what starts/stops the request intervals. I also removed the unnecessary `unsubscribeFromAll` function which only ever contained a single callback.


src/application/machines.ts
```diff
@@ -87,15 +87,14 @@ export function createDevtoolsMachine({ actions }: { actions: Actions }) {
...
-          entry: ["unsubscribeFromAll"],
+          entry: ["cancelRequestInterval"],
         },
         timedout: {},
         notFound: {
           on: {
             retry: "retrying",
             connect: "connected",
           },
-          entry: "unsubscribeFromAll",
```
jerelmiller, line 98: With this now called `cancelRequestInterval`, this action made no sense here since the only time we'll have a request interval is when we start one, which happens when we enter the `connected` state. There is no path from `connected` to `notFound`, so this will never have any effect.

src/extension/devtools/devtools.ts
```diff
@@ -45,12 +41,6 @@ const portAdapter = createPortMessageAdapter(() =>
 const clientPort = createActor<ClientMessage>(portAdapter);
 const rpcClient = createRpcClient<DevtoolsRPCMessage>(portAdapter);
 
-devtoolsMachine.subscribe(({ value }) => {
-  if (value === "connected") {
```
jerelmiller, line 49: This is captured in the `startRequestInterval` machine action, so we can remove the redundancy here.

src/extension/devtools/devtools.ts
```diff
@@ -107,13 +97,6 @@ function startRequestInterval(ms = 500) {
   return () => clearTimeout(id);
 }
 
-const unsubscribers = new Set<() => void>();
-
-function unsubscribeFromAll() {
```
jerelmiller, line 112: When the devtools cancellation functions were much more complicated, I had a dream that they would all live here to made it easy to mass-cancel all of them. We never did end up doing much with that and we were able to simplify those other cancellation functions in the meantime.

This only now ever handles a single callback (the `cancelRequestInterval` function), so I have now named it as such to make this easier to follow.

src/extension/devtools/devtools.ts
```diff
@@ -100,20 +90,11 @@ function startRequestInterval(ms = 500) {
     }
   }
 
-  if (devtoolsMachine.state.value === "connected") {
```
jerelmiller, line 103: We already check this conditional in other places that call this function, so we don't need to check it again here.

src/application/machines.ts
```diff
@@ -36,22 +34,14 @@ type State = {
...
+  cancelRequestInterval: StateMachine.ActionFunction<Context, Events>;
   startRequestInterval: StateMachine.ActionFunction<Context, Events>;
-  unsubscribeFromAll: StateMachine.ActionFunction<Context, Events>;
 };
 
 export function createDevtoolsMachine({ actions }: { actions: Actions }) {
   return createMachine<Context, Events, State>(
     {
       initial: "initialized",
-      context: {
```
jerelmiller, line 47: I'm removing `context` on the machine because we never really kept this up-to-date with the latest values from the client, which means we weren't really using this as a source-of-truth. The only time we read this context was when we initialized the panel so that we could show data quickly. That has now been updated to use the rpc client to pull this data instead.

src/application/machines.ts
```diff
@@ -1,21 +1,10 @@
-import type { QueryInfo } from "../extension/tab/helpers";
-import type { JSONObject } from "./types/json";
-
 import type { StateMachine } from "@xstate/fsm";
-import { createMachine, assign } from "@xstate/fsm";
+import { createMachine } from "@xstate/fsm";
 
-export interface ClientContext {
```
jerelmiller, line 7: This type is only now ever used in `messages.ts` so it moved over there.

src/extension/devtools/panel.ts
```diff
@@ -15,7 +15,7 @@ panelWindow.on("initializePanel", (message) => {
 panelWindow.on("devtoolsStateChanged", (message) => {
   devtoolsState(message.state);
 
-  if (message.state === "connected") {
+  if (message.state !== "connected") {
```
jerelmiller, line 18: This was added in https://github.com/apollographql/apollo-client-devtools/pull/1379/files#diff-7c1d36e9e85ad36384132856e68b4e2eaf0d11d4d64e911381c8878329a5f38aR18, but removed when we reverted that release. I'm adding it back here since we want this behavior. See my [original comment](https://github.com/apollographql/apollo-client-devtools/pull/1379/files#r1607185319) for more information on the context of this change.

phryneas (approved):
Looking good :)

PR #1413, opened 2024-06-10
Fix issue where messages sent via the `window` adapter could be echoed back to the sender

jerelmiller:
While working on the feature to work with multiple client instances on a page, I noticed that some unnecessary messages were moving between the content and devtools scripts. Digging further, I saw that the rpc requests were getting echoed back to the sender of the request.

### DIFF
--- a/.changeset/tricky-rats-hang.md
+++ b/.changeset/tricky-rats-hang.md
@@ -1,0 +1,5 @@
+---
+"apollo-client-devtools": patch
+---
+
+Fix issue where messages sent via the `window` adapter could be echoed back to the sender.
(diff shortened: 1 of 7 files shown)


phryneas (approved):
I think you could avoid a bit complexity with id tracking, timeouts, etc. if you instead added an `origin` property and created an `origin` id for each adapter - but either way, this looks like a valuable change, so I'll approve it like it is :)

jerelmiller:
@phryneas I found a case where that doesn't work due to how we send messages between the devtools script and the panel script, so it completely breaks devtools. I'll avoid the long explanation unless you're curious, but for now I'll leave this as-is. I've spotted some more opportunities for reducing the message load and I'll consider swapping out this implementation then.

PR #1418, opened 2024-06-14
Add support for inspecting multiple clients

### LINKED ISSUE #822: Apollo dev tools 4.0 not showing any data with multiple Apollo Clients
**Intended outcome:**
I expect to be able to see the queries data in browser extension.

**Actual outcome:**
Not showing any data.

Install extension in browser.

 - OS: Linux Manjaro
 - Browser chromium, firefox
 - Browser version: Firefox Developer Edition 98, Chromium Version 98.0.4758.102 (Official Build) Manjaro Linux (64-bit)
 - Extension version 4.0

The dev tools version is 4, same situation in Chrome as in Firefox. "@apollo/client": "^3.5.7",
```
const createApolloClient = (authToken) => {
  return new ApolloClient({
    link: new HttpLink({
      uri: "https://...url",
      headers: {
        Authorization: `Bearer ${authToken}`,
      },
    }),
    cache: new InMemoryCache(),
    connectToDevTools: true, // should not be necessary but added just in case
    defaultOptions: {
      watchQuery: {
        fetchPolicy: "cache-and-network",
        pollInterval: 3000,
      },
    },
  });
};
```

The application is React + Vitejs, if it makes any difference... No idea where to start any debuging, please help...

jerelmiller:
Fixes #822

Adds support for connecting and inspecting multiple Apollo Client instances simultaneously. This allows apps that use multiple client instances (for example with a microfrontends) to inspect the state of each of these client instances. This takes advantage of the upcoming feature in 3.11 that will enable users to [name their client instance](https://github.com/apollographql/apollo-client/pull/11936) making it easy to identify which client they're inspecting.

This works with the "Explorer" tab as well. Changing clients will rerun the introspection query for that client and send GraphQL requests through that client.

Screenshot 2024-07-11 at 6 02 19 PM
Screenshot 2024-07-11 at 6 02 24 PM

---

This PR includes a major refactor to several areas of the code base. Up to this point, much of the logic for handling messages from the hook script was handled in the `devtools.ts` script. That script was then responsible for syncing data to the React app running the UI. The UI used its own `ApolloClient` instance to sync data to the UIs, but it did so exclusively through manual cache updates in combination with the `@client` directive. 

This PR switches to use a `SchemaLink` in combination with RPC calls to treat it more like client/server setup. This means we can move things like data polling to the React app. This has the added benefit that we only need to sync the data requested by a particular view. For example, we know longer need to sync all mutation data and cache data when looking at the "Queries" tab, which should be more efficient and add less strain on the running app. This is a natural result of switching to the resolver model for syncing that data.

The devtools script is now not much more than a message bus between the client port and the panel window, which drastically simplifies it. Moving forward, we'll add much more of the logic-heavy code to the React app itself, making it feel more like a traditional SPA; a pattern we are all familiar with.


src/application/components/Cache/Cache.tsx
```diff
@@ -51,15 +52,22 @@ function filterCache(cache: Cache, searchTerm: string) {
...
-    [data?.cache]
-  );
+  const { loading, data } = useQuery(GET_CACHE, {
+    variables: { id: clientId as string },
+    skip: clientId == null,
+    pollInterval: 1000,
+    fetchPolicy: "cache-and-network",
+  });
+
+  const cache = useMemo(() => data?.client?.cache ?? {}, [data?.client.cache]);
```
phryneas, line 70:
```suggestion
  const cache = data?.client?.cache ?? STABLE_EMPTY_OBJ
```

src/application/App.tsx
```diff
@@ -106,14 +120,33 @@ ${SECTIONS.devtoolsVersion}
...
 
-  const clientVersion = data?.clientVersion;
+  const { data: clientData } = useQuery(CLIENT_QUERY, {
+    variables: { id: selectedClientId as string },
+    skip: !selectedClientId,
+    pollInterval: 1000,
+  });
+
+  const client = clientData?.client;
+  const clientVersion = clientData?.client.version;
```
phryneas, line 140:
```suggestion
  const clientVersion = client?.version;
```

src/extension/tab/hook.ts
```diff
@@ -54,27 +55,39 @@ const tab = createWindowActor<ClientMessage>(window);
...
+      ? // @ts-expect-error Apollo Client 3.0 - 3.2
+        ac.queryManager.mutationStore?.getStore()
+      : // Apollo Client 3.3
+        ac?.queryManager.mutationStore) ?? {}
+  );
+}
+
+// Keep a reverse mapping of client -> id to ensure we don't register the same
+// client multiple times.
+const knownClients = new Map<ApolloClient<SafeAny>, string>();
```
phryneas, line 82:
```suggestion
const knownClients = new WeakMap<ApolloClient<SafeAny>, string>();
```

Let's ensure that the devtools don't cause any memory leaks

phryneas (commented):
Only very minor nits so far.

src/extension/tab/hook.ts
jerelmiller: Unfortunately I need to be able to traverse this map to be able to look up which client belongs to which `id`, otherwise it defeats the purpose of this feature.

I probably could use some combination of `WeakRef` and `WeakSet`/`WeakMap` for this to be able to perform some kind of iteration along with `FinalizationRegistry` to detect when the client has been garbage collected, but definitely a bit more complicated than it seems. How do you feel about a followup PR that does this in isolation?

phryneas (dismissed):
This seems to have problems with older AC versions - the dropdown doesn't show up. Once that's fixed, LGTM

jerelmiller:
Discussed over Slack, but posting here for transparency:

The issue with the older client versions wasn't a bug with the devtools, but rather because the client config updated in this PR is using the newer `devtools` option that will be introduced in 3.11 instead of the now-deprecated `connectToDevtools` option. The `connectToDevtools` option, when not set explicitly, will only be set to `true` if it doesn't detect a `window.__APOLLO_CLIENT__` instance: https://github.com/apollographql/apollo-client/blob/3f6d023db2d7d6bf977060dd5db6291fd64e85e6/src/core/ApolloClient.ts#L204-L206

After downgrading and reverting back to the explicit `connectToDevtools: true` option, multiple clients started showing up again in the devtools.

src/application/components/Cache/Cache.tsx
```diff
@@ -51,15 +53,27 @@ function filterCache(cache: Cache, searchTerm: string) {
...
-  );
+  const { loading, data, startPolling, stopPolling } = useQuery(GET_CACHE, {
+    variables: { id: clientId as string },
+    skip: clientId == null,
+    pollInterval: 500,
+    fetchPolicy: "cache-and-network",
+  });
+
+  useActorEvent("panelHidden", () => stopPolling());
+  useActorEvent("panelShown", () => startPolling(500));
```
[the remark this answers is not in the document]
jerelmiller: Will look to address this in a followup PR to avoid holding up this release. Thanks for the review!

PR #1427, opened 2024-06-24
Update deps

jerelmiller:
Supersedes #1411
Supersedes #1366
Supersedes #1320

Manually updates some dependencies that renovate is trying to upgrade but fails. I'm ignoring the v2 update for `react-resizable-panels` for now because this is causing some test failures that require a bit more digging into.


babel.config.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
jerelmiller, line 1: This is needed by Jest in order to transform ESM libraries in tests

jest.config.js
```diff
@@ -1,5 +1,31 @@
+const esModules = [
```
jerelmiller, line 1: This is a giant list because `react-markdown` includes a lot of ESM dependencies that need to be transformed. I wanted to use the real implementation rather than our mocked one, so this replaces our library mock that we had before. This was also needed because `react-json-tree` was updated to ESM-only in the latest release.

jest.config.js
```diff
@@ -1,5 +1,31 @@
...
+  "trough",
+  "unified",
+  "unist-.+",
+  "vfile",
+  "vfile-message",
+].join("|");
+
 export default {
-  preset: "ts-jest",
+  preset: "ts-jest/presets/js-with-babel",
```
jerelmiller, line 28: This is needed to transform `.js` files in `node_modules` to commonjs syntax in order for tests to run properly. The `ts-jest` preset on its own did not transform `.js` files, so our tests choked when importing from ESM libraries.

jest.config.js
```diff
@@ -11,11 +37,11 @@ export default {
     "/node_modules/",
     "/development/",
   ],
+  transformIgnorePatterns: [`/node_modules/(?!${esModules})`],
```
phryneas, line 40:
```suggestion
  transformIgnorePatterns: [`/node_modules/(?!${esModules})/`],
```

Maybe we should ensure that this is not part of a longer folder name?

phryneas (approved):

jest.config.js
jerelmiller: Works for me!

PR #1438, opened 2024-07-01
Use `queryId` as `id` for queries

jerelmiller:
Switches the value of `id` used for queries to the `queryId` its associated with in Apollo Client rather than its index in the map. This should allow us to build additional functionality in the future that targets the proper query rather than guessing based on its index in the map.

### DIFF
--- a/src/extension/tab/helpers.ts
+++ b/src/extension/tab/helpers.ts
@@ -44,6 +44,7 @@ export interface SerializedError {
 }
 
 export type QueryDetails = {
+  id: string;
   document: DocumentNode;
   variables?: Variables;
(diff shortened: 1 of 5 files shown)


src/extension/tab/helpers.ts
```diff
@@ -46,6 +46,7 @@ export interface SerializedError {
 }
 
 export type QueryInfo = {
```
phryneas, line 48: Could we rename this type at some point? It's a bit of a red herring since we have a type of the same name over in AC.

phryneas (approved):
Looks good to me :thumbsup:

PR #1454, opened 2024-07-23
Fix issue where data would disappear after refreshing the page

### LINKED ISSUE #1452: Data seems to disappear after refreshing the page
After loading the devtools for the first time, refreshing the page seems to clear the data after a few second delay. Refreshing the page after this does not seem to fix the issue.

This only seems to occur after a full page refresh. If I navigate the app, the queries and mutations update just fine.

### `@apollo/client` version
3.11.0

### Apollo Client Devtools version
4.17.0

jerelmiller:
Fixes #1452

#1418 did a bunch of refactoring work to reorganize how and where data was requested for the devtools. As part of that change, a GraphQL-like "server" was spun up through `SchemaLink` that used RPC calls to resolve data.

After refreshing the page, we'd put the devtools in a disconnected state and in doing so, would call `client.resetStore`. `resetStore` causes active queries to be refetched, and in doing so, we'd see RPC request timeouts in cases where the disconnect was caused by reloading the page since the content scripts had not yet been injected and could not receive the RPC request message. This timeout would cause the query to get put into an error mode, which would clear the query data and cause the UI to show nothing in the UI.

This fix changes to use `clearStore` instead of `resetStore` to prevent the RPC calls from firing when the content script was not yet ready to receive messages and instead will refetch the list of clients after the first client is registered again.

---

Ultimately I'd love to introduce a message queue so that we can queue up messages from either end of the devtools in the event the port is disconnected. Upon connecting, we can send queued messages through the queue to process them. This requires a bit more thought and work, so this fix should be sufficient for now.

### DIFF
--- a/src/application/App.tsx
+++ b/src/application/App.tsx
@@ -74,7 +74,7 @@ export const App = () => {
     devtoolsMachine.provide({
       actions: {
         resetStore: () => {
-          apolloClient.resetStore().catch(() => {});
+          apolloClient.clearStore().catch(() => {});
         },
       },
     })
--- a/src/application/App.tsx
+++ b/src/application/App.tsx
@@ -36,7 +36,7 @@ import { isSnapshotRelease, parseSnapshotRelease } from "./utilities/github";
 import { Select } from "./components/Select";
 import { Divider } from "./components/Divider";
 import { useActorEvent } from "./hooks/useActorEvent";
-import { addClient, removeClient } from ".";
+import { removeClient } from ".";
 
 const APP_QUERY: TypedDocumentNode<AppQuery, AppQueryVariables> = gql`
   query AppQuery {
--- a/src/application/App.tsx
+++ b/src/application/App.tsx
@@ -79,17 +79,23 @@ export const App = () => {
       },
     })
   );
-  const { data, client: apolloClient } = useQuery(APP_QUERY, {
-    errorPolicy: "all",
-  });
+  const {
+    data,
+    client: apolloClient,
+    refetch,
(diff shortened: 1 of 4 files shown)


src/application/App.tsx
```diff
@@ -74,22 +74,28 @@ export const App = () => {
     devtoolsMachine.provide({
       actions: {
         resetStore: () => {
-          apolloClient.resetStore().catch(() => {});
+          apolloClient.clearStore().catch(() => {});
```
jerelmiller, line 77: Because this now uses `clearStore` instead of `resetStore`, active queries no longer clear out their data. This introduces a small regression where refreshing the page will show the stale data in the app until a new client is registered.

Unfortunately I've been unable to get the mounted queries to respond to cache updates after the cache has been cleared, so the data remains on screen. The `refetch` after the first client is registered will fix this issue.

src/application/App.tsx
```diff
@@ -74,22 +74,28 @@ export const App = () => {
...
+  } = useQuery(APP_QUERY, { errorPolicy: "all" });
 
   useActorEvent("connectToDevtools", () => {
     send({ type: "connect" });
   });
 
-  useActorEvent("registerClient", (message) => {
+  useActorEvent("registerClient", () => {
     send({ type: "connect" });
-    addClient(message.payload);
```
jerelmiller, line 92: I updated `addClient` to use `client.writeQuery` to run a `broadcastQueries`, but unfortunately that didn't seem to have an effect on the mounted queries after `clearStore` was run.

This is definitely something we should look into fixing in the client as its easy for `useQuery` to get out-of-sync with the cache after a `clearStore` has been called.

PR #1460, opened 2024-07-26
Remove unneeded initialization message + fix stale data issue on disconnect

jerelmiller:
I found an opportunity to remove the unneeded `connectToDevtools` and `connectToClient` messages since we are now fetching the list of clients when opening devtools for the first time. We can use the result of this query to determine if we are connected by simply seeing if we get back at least 1 client.

This PR also fixes the issue where disconnecting from a client that was previously connected would show its stale data on screen until another client connected to the devtools. This is done by refetching the list of clients after we clear the store to ensure the `useQuery` hook is refreshed with the latest data.

### DIFF
--- a/src/application/App.tsx
+++ b/src/application/App.tsx
@@ -89,10 +89,6 @@ export const App = () => {
     refetch,
   } = useQuery(APP_QUERY, { errorPolicy: "all" });
 
-  useActorEvent("connectToDevtools", () => {
-    send({ type: "connect" });
-  });
-
   useActorEvent("registerClient", () => {
     send({ type: "connect" });
     // Unfortunately after we clear the store above, the query ends up "stuck"
--- a/src/application/App.tsx
+++ b/src/application/App.tsx
@@ -148,7 +148,7 @@ export const App = () => {
     if (clients.length) {
       send({ type: "connect" });
     }
-  }, [send, clients]);
+  }, [send, clients.length]);
 
   return (
     <>
--- a/src/application/App.tsx
+++ b/src/application/App.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import type { TypedDocumentNode } from "@apollo/client";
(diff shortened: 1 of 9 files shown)


src/application/components/BannerAlert.tsx
```diff
@@ -37,7 +37,7 @@ export function BannerAlert() {
             }
           )}
         >
-          {banner.type === "loading" && <LoadingSpinner size="2xsmall" />}
+          {banner.type === "loading" && <Spinner size="xs" />}
```
jerelmiller, line 40: I took this opportunity to swap out the loading spinner for the orbit themed spinner

src/application/machines/devtoolsMachine.tsx
```diff
@@ -23,19 +22,17 @@ export const devtoolsMachine = setup({
   actions: {
     openModal: assign({ modalOpen: true }),
     closeModal: assign({ modalOpen: false }),
-    connectToClient: () => {
+    notifyWaitingForConnection: () => {
       BannerAlert.show({
         type: "loading",
-        content: "Looking for client...",
+        content: "Waiting for client to connect...",
```
jerelmiller, line 28: I updated these messages to be a bit more accurate as to what is happening with the more modern versions of devtools

src/application/App.tsx
```diff
@@ -144,6 +144,12 @@ export const App = () => {
     setSelectedClientId(clientIds[0]);
   }
 
+  useEffect(() => {
+    if (clients.length) {
+      send({ type: "connect" });
+    }
+  }, [send, clients]);
```
phryneas, line 151:
```suggestion
  }, [send, clients.length]);
```

No need to rely on referential equality here when all we rely on is the length, right?

phryneas (approved):
LGTM :)

phryneas (approved):

src/application/App.tsx
jerelmiller: Good call!

PR #1461, opened 2024-07-26
Create single message bridge and simplify message types

jerelmiller:
Currently messages sent through the `actor` have extra steps needed to ensure the message makes it safely between the `hook` and the devtools app. Adding a new message requires adding the message type to the `messages` file, then remembering to add `forward` calls in both the `tab.ts` and `devtools.ts` files in order for the message to relay between the different message adapters.

This PR aims to simplify this by creating a unified message bridge that handles forwarding Apollo Client Devtools messages between any two actors in any direction. This is a replacement for the rpc message bridge that existed but only handled rpc messages. This change means that adding support for a new message can be done in a single place by editing the message types to propagate everywhere.

### DIFF
--- a/src/extension/__tests__/actor.test.ts
+++ b/src/extension/__tests__/actor.test.ts
@@ -222,38 +222,3 @@ test("re-adds listener on adapter when subscribing actor listener after disconne
   actor.on("connect", handleConnect);
   expect(adapter.addListener).toHaveBeenCalledTimes(1);
 });
-
-test("forwards messages to another actor", () => {
-  type Message = { type: "connect"; payload: string } | { type: "disconnect" };
-
-  const proxyAdapter = createTestAdapter<Message>();
-  const actorAdapter = createTestAdapter<Message>();
-  const proxy = createActor<Message>(proxyAdapter);
-  const actor = createActor<Message>(actorAdapter);
-
-  proxy.forward("connect", actor);
-  proxy.forward("disconnect", actor);
-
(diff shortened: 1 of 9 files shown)


src/extension/__tests__/actor.test.ts
```diff
@@ -29,171 +29,154 @@ function createTestAdapter<Messages = unknown>() {
 }
 
 test("sends messages to specified adapter in devtools message format", () => {
-  type Message = { type: "test"; payload: string };
```
jerelmiller, line 32: The one downside to this change is that we lose the ability to do generic messages in tests and now must use the real message types. This may be a bit more brittle over time as message types are refactored, but I think its a fair tradeoff.

src/extension/tab/tab.ts
```diff
@@ -1,30 +1,18 @@
...
-
-createRPCBridge(portAdapter, createWindowMessageAdapter(window));
-
-devtools.forward("explorerSubscriptionTermination", tab);
-devtools.forward("explorerRequest", tab);
-
-tab.forward("registerClient", devtools);
-tab.forward("clientTerminated", devtools);
-tab.forward("explorerResponse", devtools);
+createMessageBridge(portAdapter, createWindowMessageAdapter(window));
```
jerelmiller, line 15: Lost in all the type refactoring is the biggest change which is the simplification of the message transfer between two actors. No more need for forwarding specific messages!

jerelmiller:
I'm going to split this up into 2 PRs. Moving this to draft for now.

phryneas (approved):
I dig it!

PR #1462, opened 2024-07-26
Unify and simplify actor and RPC message types

jerelmiller:
I saw an opportunity to simplify the message types in the app. We use a ton of generics with both the actor and RPC models, but looking through this more, this seemed largely unnecessary because we don't have messages going a bunch of different ways to different things. We largely want to create a single actor/rpc client and use that for messaging in a specific area of the extension.

As a part of this, I decided to simplify the types and remove the need for generics. I've combined the message types together into a single union type that has now moved to the respective files where the implementation lives. This should make it easier to know both where to put the type and avoid confusion on what message type should include new types.

### DIFF
--- a/src/extension/actor.ts
+++ b/src/extension/actor.ts
@@ -18,7 +18,7 @@ export function createActor<
   Messages extends MessageFormat = {
     type: "Error: Pass <Messages> to `createActor<Messages>()`";
   },
->(adapter: MessageAdapter<Messages>): Actor<Messages> {
+>(adapter: MessageAdapter): Actor<Messages> {
   let removeListener: (() => void) | null = null;
   const messageListeners = new Map<
     Messages["type"],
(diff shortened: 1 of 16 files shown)


.changeset/six-months-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"apollo-client-devtools": patch
+---
+
+Create a universal message bridge to handle both actor and rpc messages.
```
jerelmiller, line 5: Need to move this to the other PR

src/extension/actor.ts
```diff
@@ -57,12 +77,15 @@ export function createActor<
...
     let listeners = messageListeners.get(name) as Set<typeof callback>;
 
     if (!listeners) {
       listeners = new Set();
-      messageListeners.set(name, listeners as Set<(message: Messages) => void>);
+      messageListeners.set(
+        name,
+        listeners as Set<(message: ActorMessage) => void>
+      );
     }
```
phryneas, line 89:
```suggestion
    if (!listeners) {
      listeners = new Set();
      messageListeners.set(name, listeners);
    }
```

Seems the cast is not neccessary here

phryneas (approved):
Looks like a good improvement :)

src/extension/actor.ts
jerelmiller: Good catch!

PR #1497, opened 2024-08-29
DevTools <-> VSCode integration

phryneas:
This is in *very* early stages and split up between here and https://github.com/apollographql/vscode-graphql/pull/188


phryneas:
And the Explorer tab is working :tada:

image

phryneas:
I'll fork off here into additional features - this is a good size for a review already :)

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+      signal.addEventListener("abort", () => cleanupFn(signal.reason), {
+        once: true,
+      });
+    }
+  }
+  return {
+    cleanup: (reason: Reason = "UNREGISTER") => cleanup(reason),
+    signal,
+    registerCleanup,
+  };
```
jerelmiller, line 161: Since this function is called `getCleanupController`, I'd expect to see the `AbortController` returned as a property here. Any reason we couldn't replace both `cleanup` and `signal` with `controller`? Is the `cleanup` function purely for type safety?

```suggestion
  return {
    controller: cleanupController,
    registerCleanup,
  };
```

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+  function getClient() {
+    const client = clientRef.deref();
+    if (client) {
+      return getPrivateAccess(client);
+    } else {
+      cleanup("CLIENT_GC");
+    }
+  }
+
+  const ws = new WebSocket(url);
```
jerelmiller, line 63: I'm curious how multiple clients work with separate socket connections. While I think this is fine for outgoing messages (i.e. from the client -> devtools), does creating a separate connection per client mean that the devtools sends messages to every connected socket?

If so, I think we need to be careful that the rpc message handlers aren't all responding to the same message, otherwise whichever handler responds first will be the data displayed in the devtools. That handler may not necessarily be the client that is selected in the dropdown.

The current `hook` script used for the browser devtools is deliberate about using client IDs for this purpose so that we ensure we're always responding with data for the currently selected client.

Can you double check that each of these handlers don't respond with data for multiple clients? This might be a pretty big bug otherwise (at least for thsoe using multiple clients).

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+    },
+  };
+}
+
+function getCleanupController() {
+  const cleanupContoller = new AbortController();
+  const cleanup = cleanupContoller.abort.bind(cleanupContoller);
+  const signal = cleanupContoller.signal;
+  setMaxListeners(20, signal);
+  function registerCleanup(
```
jerelmiller, line 145:
```suggestion
  function onCleanup(
```

Maybe just me, but I kinda like calling this `on*` since this is basically just storing event listeners. Thoughts?

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+    "open",
+    function open() {
+      const client = getClient();
+      if (!client) {
+        return;
+      }
+      wsActor.send({
+        type: "registerClient",
+        payload: {
+          id: "",
```
jerelmiller, line 103: We will need to create a unique ID per connected client. The select dropdown [uses this `id`](https://github.com/apollographql/apollo-client-devtools/blob/a413e0b8fe2c3ccab7225950439bbf6e9dccab62/src/application/App.tsx#L226) to create separate entries in the dropdown. If all clients use the same `id`, we won't be able to properly display the options in the dropdown.

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+  return {
+    connected: new Promise<void>((resolve) =>
+      ws.addEventListener("open", () => resolve(), { once: true, signal })
+    ),
+    unregister: cleanup,
+    onCleanup: registerCleanup,
+  };
+}
+
+function createWsAdapter(ws: WebSocket, signal: AbortSignal): MessageAdapter {
```
jerelmiller, line 125: I think it would be good to do a bit of organization on the existing code so that this is near the other code, but ok for now and we can refactor in a future PR.

src/extension/vscode/server.ts
```diff
@@ -0,0 +1,110 @@
...
+    string,
+    {
+      rpcClient: ReturnType<typeof createRpcClient>;
+      actor: Actor;
+      info: ApolloClientInfo;
+    }
+  >();
+
+  const messageAdapter = adapter as MessageAdapter;
+  const tab = createActor(messageAdapter);
```
jerelmiller, line 32: [nit] No need to call this `tab` since we're in a different context (unless you prefer this). I think the browser extension calls it this since it represents the code running in the user's browser tab, but that might not make much sense here.

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+      });
+      loadErrorCodes(wsRpcClient, client.version);
+    },
+    { once: true, signal }
+  );
+
+  return {
+    connected: new Promise<void>((resolve) =>
+      ws.addEventListener("open", () => resolve(), { once: true, signal })
+    ),
```
jerelmiller, line 119: I'm not a huge fan of this name mostly because reading this I keep expecting this to be a boolean, but its a promise instead.

Out of curiosity, could we just use a callback here and call this `onConnected`? We can remove the promise altogether then. 

```suggestion
    onConnected: (callback: () => void) => {
      ws.addEventListener("open", () => callback(), { once: true, signal })
    },
```

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+    unregister: cleanup,
+    onCleanup: registerCleanup,
```
jerelmiller, line 121:
```suggestion
    disconnect: cleanup,
    onDisconnect: registerCleanup,
```

What about calling this `disconnect`? I think this will probably be a bit more intuitive than "cleanup" since its not immediately clear what is being cleaned up without knowing the implementation details. Thoughts?

src/extension/vscode/server.ts
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  server.on("connection", function connection(ws) {
+    const id = createId();
+    const wsAdapter = createSocketMessageAdapter(ws);
+    const wsRpcClient = createRpcClient(wsAdapter);
+    const wsRpcHandler = createRpcHandler(wsAdapter);
+    const wsActor = createActor(wsAdapter);
+    wsActor.on("registerClient", ({ payload }) => {
+      payload.id = id;
```
jerelmiller, line 56: Ah ok I see this is set here instead of in `client.ts`, which I missed when adding the other comment. 

I'm curious why you decided to go this route of creating a handler/rpc client per apollo client instance? If you're able to use a single client/rpc handler, you would then be able to take advantage of the [message bridge](https://github.com/apollographql/apollo-client-devtools/blob/a413e0b8fe2c3ccab7225950439bbf6e9dccab62/src/extension/messageAdapters.ts#L104-L124) since this script would essentially just be the middle man between the client script and devtools.

I'd recommend taking a stab at that, otherwise we'll need to edit this file anytime we want to add a new message from devtools to the client (such as the memory panel that will be added soon). I used to do this in `tab.ts` and it there were definitely times that I scratched my head wondering why something wasn't working only to realize I forgot to add the message forwarding between the two. The message bridge definitely helps out since its a bit more automatic.

jerelmiller (changes requested):
I might have more with fresh eyes, but I think this is a good set of feedback to take a look at for now. I'd definitely recommend seeing if you can reduce to a single rpc client/message adapter in the server if possible so you can take advantage of the message bridge.

package.vscode.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@apollo/client-devtools-vscode",
+  "version": "4.18.4-pre.0",
```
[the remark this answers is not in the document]
phryneas: That was just for my local build back when I forked off - when we publish the package, it will go through changesets and inherit the version of the extension :)

src/extension/vscode/client.ts
```diff
@@ -0,0 +1,162 @@
...
+  | "WS_DISCONNECTED"
+  | "CLIENT_GC"
+  | "WS_ERROR"
+  | "UNREGISTER"
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | (string & {});
+
+export function registerClient(
+  client: ApolloClient<any>,
+  url: string | URL
```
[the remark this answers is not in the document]
phryneas: > Should we have a default value here of ws://localhost:7095? Is it safe to assume that the majority of people will probably use the default VSCode configuration locally?

I would actually assume that for most people it would be a tunnel address - I wouldn't want to hide this with a default here.

src/extension/vscode/client.ts
phryneas: All the mapping "which client id references which connection" happens on the server - this client would never get messages meant for other clients. But I'll double-check.

src/extension/vscode/client.ts
phryneas: I know. Client ids are assigned in the server, which is the only instance that has the knowledge which client ids are already taken and which client id refers to which connection.

src/extension/vscode/client.ts
phryneas: That would miss the event if you added an `onConnected` event after the connection is already established and handling *that* would add extra logic. I'm open to changing the name, but I'd like to keep the promise, as that handles that nicely.

src/extension/vscode/client.ts
phryneas: It's called `unregister` right now, because the function itself is named `registerClient`.

We could rename `registerClient` to `connectClientToDevTools` or something, then `disconnect` would make sense :thinking:
Not sure if I like that better, though - WDYT?

src/extension/vscode/client.ts
phryneas: The `cleanup` function exists to set a default parameter to `.abort` - I could name it `.abort`, but it doesn't really describe what it does:

It cleans up everything. Even if the connection has already been closed (so we have partially aborted, so to speak), it cleans up all the other stuff. It's not only called to abort the action, but to also clean up in the case something was already aborted externally.

src/extension/vscode/polyfills.ts
```diff
@@ -0,0 +1,39 @@
...
+        } satisfies Omit<FinalizationRegistry<T>, typeof Symbol.toStringTag>;
+      } as any as typeof FinalizationRegistry);
+
+export { _FinalizationRegistry as FinalizationRegistry };
+
+/**
+ * A workaround to set the `maxListeners` property of a node EventEmitter without having to import
+ * the `node:events` module, which would make the code non-portable.
+ */
+export function setMaxListeners(maxListeners: number, emitter: any) {
```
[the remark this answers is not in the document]
phryneas: This imitates the method signature of `setMaxListeners` from `node:events`:
https://nodejs.org/api/events.html#eventssetmaxlistenersn-eventtargets

src/extension/vscode/server.ts
phryneas: The message bridge would work if we had the guarantee of a single connection, but you can totally connect two different processes to one server (e.g. connect an iOS emulator and an Android emulator at the same time). So either at that point I send everything everywhere, or I have to keep track of who is who in the server - I went for the latter.

Also, we could in the future add some "snapshot" functionality here inside the server that would work by creating a "virtual client" that would be fully controlled by the server.

package.vscode.json
```diff
@@ -0,0 +1,36 @@
...
+    "vscode-client.js",
+    "vscode-server.js",
+    "panel.js",
+    "*.d.ts",
+    "package.json",
+    "LICENSE.md",
+    "README.md"
+  ],
+  "peerDependencies": {
+    "@apollo/client": "^3.10.4"
```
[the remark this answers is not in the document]
phryneas: I think we could even go deeper, until 3.7 or something. Good point!

src/extension/vscode/polyfills.ts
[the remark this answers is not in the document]
jerelmiller: TIL

src/extension/vscode/client.ts
jerelmiller: Actually I do like that better! I had a similar thought this morning before reading your comment so I'm on board. A name like `connectClienttoDevTools` makes it immediately obvious what that block of code does without having to follow the import to understand what you're registering your client with. On top of that, it better follows with our `connectToDevTools` option naming we've had in Apollo Client for for a while so theres some symmetry there 🙂

phryneas: Then I'd suggest we make it *really* obvious with a `connectApolloClientToVSCodeDevtools`? :)

jerelmiller: Haha either one, but something with `connect` and `devTools` in it makes sense to me :)

src/extension/vscode/client.ts
jerelmiller: Thats a good point. Let's see if we can rename this then. 

Would it be out of the question for `registerClient` (or whatever the new name will be) to return a promise instead? If not, my brain is out of names 😆

package.vscode.json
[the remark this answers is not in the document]
phryneas: v3.4.0-rc.4 should actually be enough, I'm putting 3.4.0.

src/extension/vscode/polyfills.ts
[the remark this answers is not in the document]
phryneas: Yeah, me too ^^

src/extension/vscode/client.ts
phryneas: done

src/extension/vscode/client.ts
phryneas: created on the client now

src/extension/vscode/client.ts
phryneas: `connectedPromise` - will at least not be confused with a boolean, and also rejects now when there is a connection error

src/extension/vscode/client.ts
phryneas: renamed to `connectApolloClientToVSCodeDevTools` and also edited the shape of the returned object

src/extension/vscode/server.ts
phryneas: Creating the `id` in`registerClient` now, I just hope there won't be any collisions in low-entropy environments

apollo-bot2:
## Detected SAST Vulnerabilities
## :red_circle: Potential Security Issues Found

If you are seeing this message, it means that [the security scanning tool](https://semgrep.dev/) that Apollo uses to test our source has identified a potential security issue in code introduced or updated via your branch. Information about what was detected and steps to move forward are below.

If the scanner detected a legitimate issue, please take action to correct it prior to merging this PR. The action required will vary based on the detection. If the detection is a false positive, please follow the steps below to resolve the issue.

###  Issue Detected 
| Message | Caused CI Failure? | Link | Path | Signature |
|---------|--------------------|------|------|-----------|
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/apollo-client-devtools/blob/5bc7aa4f50fde50bbbb3185a43281956054ef3d2/src/extension/vscode/client.ts?plain=1#L45" target="_blank">Link</a> | src/extension/vscode/client.ts | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
<hr />
<details>
    <summary><h3>False Positive Resolution Process</h3></summary>

The easiest way to resolve a false positive is to add a comment containing `nosemgrep` to the code that triggered the detection. You can add the comment directly to the line triggering the detection or on the line immediately above it. For example:

```python
#!/usr/bin/env python3

def my_function():
    my_code_generating_detection() # nosemgrep
```

<details>
    <summary><h4>What if I can't add a line comment?</h4></summary>

If you can't add a comment because the detection is generated by a file that does not support comments (like JSON), you can use Semgrep's `.semgrepignore` file. For more information about how to create an exclusion via `.semgrepignore`, refer to the [Semgrep documentation](https://semgrep.dev/docs/ignoring-files-folders-code/#defining-ignored-files-and-folders-in-semgrepignore).
</details>

</details>
<hr />
<details> 
    <summary><h3>Getting Help</h3></summary>


The Apollo Security team is available to assist in resolving this issue. Please tag us on this PR using  `@apollographql/security` if you need assistance! 
</details>
<hr />
<details>
    <summary><h3>How do I know I fixed this correctly?</h3></summary>
Once you have resolved the issue completely, this message will disappear! If you're still seeing this message, there is more to do prior to merging.
</details>

src/extension/vscode/server.ts
jerelmiller: I don't know that I'm that worried to be honest. At most we'll have _maybe_ 10s of clients connected at once, at least in the 99% of cases. If we get reports of collisions between connected clients because we start getting tons of them, we can always revisit.

jerelmiller (approved):
This looks great! Thanks for making those changes :100:

PR #1513, opened 2024-09-09
Rework reconnect logic, add "Port could not be opened" modal

phryneas:
Also see the accompanying https://github.com/apollographql/vscode-graphql/pull/197


apollo-bot2:
## Detected SAST Vulnerabilities
## :red_circle: Potential Security Issues Found

If you are seeing this message, it means that [the security scanning tool](https://semgrep.dev/) that Apollo uses to test our source has identified a potential security issue in code introduced or updated via your branch. Information about what was detected and steps to move forward are below.

If the scanner detected a legitimate issue, please take action to correct it prior to merging this PR. The action required will vary based on the detection. If the detection is a false positive, please follow the steps below to resolve the issue.

###  Issues Detected 
| Message | Caused CI Failure? | Link | Path | Signature |
|---------|--------------------|------|------|-----------|
| Insecure WebSocket Detected. WebSocket Secure (wss) should be used for all WebSocket connections. | Yes | <a href="https://github.com/apollographql/apollo-client-devtools/blob/2648fde69cd6f5d59a98aaf2fd29fa4b551bf934/src/application/components/Modals/ClientNotFoundModal.vscode.tsx?plain=1#L59" target="_blank">Link</a> | src/application/components/Modals/C  lientNotFoundModal.vscode.tsx | rules.providers.semgrep.security.ja  vascript.lang.security.detect-  insecure-websocket |
| To ensure reproducible and deterministic builds, use `npm ci` rather than `npm install` in scripts. This will use the lockfile rather than updating it. | No | <a href="https://github.com/apollographql/apollo-client-devtools/blob/2648fde69cd6f5d59a98aaf2fd29fa4b551bf934/src/application/components/Modals/ClientNotFoundModal.vscode.tsx?plain=1#L37" target="_blank">Link</a> | src/application/components/Modals/C  lientNotFoundModal.vscode.tsx | rules.providers.semgrep.security.ge  neric.ci.security.use-frozen-  lockfile-npm |
<hr />
<details>
    <summary><h3>False Positive Resolution Process</h3></summary>

The easiest way to resolve a false positive is to add a comment containing `nosemgrep` to the code that triggered the detection. You can add the comment directly to the line triggering the detection or on the line immediately above it. For example:

```python
#!/usr/bin/env python3

def my_function():
    my_code_generating_detection() # nosemgrep
```

<details>
    <summary><h4>What if I can't add a line comment?</h4></summary>

If you can't add a comment because the detection is generated by a file that does not support comments (like JSON), you can use Semgrep's `.semgrepignore` file. For more information about how to create an exclusion via `.semgrepignore`, refer to the [Semgrep documentation](https://semgrep.dev/docs/ignoring-files-folders-code/#defining-ignored-files-and-folders-in-semgrepignore).
</details>

</details>
<hr />
<details> 
    <summary><h3>Getting Help</h3></summary>


The Apollo Security team is available to assist in resolving this issue. Please tag us on this PR using  `@apollographql/security` if you need assistance! 
</details>
<hr />
<details>
    <summary><h3>How do I know I fixed this correctly?</h3></summary>
Once you have resolved the issue completely, this message will disappear! If you're still seeing this message, there is more to do prior to merging.
</details>

src/application/components/Modals/ClientNotFoundModal.tsx
```diff
@@ -1,35 +1,33 @@
...
+import { ExternalLink } from "../ExternalLink";
+import { GitHubIssueLink, SECTIONS, LABELS } from "../GitHubIssueLink";
+import { Modal } from "../Modal";
 import IconGitHubSolid from "@apollo/icons/small/IconGitHubSolid.svg";
 
 interface ClientNotFoundModalProps {
   open: boolean;
-  onClose: () => void;
   onRetry: () => void;
 }
```
phryneas, line 11: This modal cannot be closed anymore - before, closing it would leave the user with a somewhat defunct UI, so it never made much sense to me. Going this step massively simplifies logic.

src/application/components/Modals/Modals.tsx
```diff
@@ -0,0 +1,54 @@
...
+    (state) => state.context.listening === false
+  );
+
+  const reconnectActor = useDevToolsSelector(
+    (state) => state.children.reconnect
+  );
+  const openClientNotFoundModal = useSelector(
+    reconnectActor,
+    (state) => state?.status === "active" && state.value === "notFound"
+  );
```
phryneas, line 24: This is how you select state from a child-machine.

src/application/machines/devtoolsMachine.tsx
```diff
@@ -1,143 +1,236 @@
...
+            guard: ({ context }) => context.registeredClients === 0,
+            target: "disconnected",
+          },
+        },
+        disconnected: {
+          entry: "notifyDisconnected",
+          invoke: {
+            id: "reconnect",
+            src: "reconnect",
+          },
```
phryneas, line 195: This invokes the child machine while in this state and stops it as soon as this state is left.

src/application/machines/devtoolsMachine.tsx
```diff
@@ -1,143 +1,233 @@
...
+            target: "connected",
+          },
+        },
       },
-      entry: ["notifyNotFound", "openModal"],
     },
   },
 });
+
+export const DevToolsMachineContext = createContext<DevToolsActor | null>(null);
```
phryneas, line 206: We now use Context to pass this down into the React app - but not the `createContext` from xState, as that would create the actor from the machine and we want to create the actor outside of React (the Actor renders React).

src/application/machines/reconnectMachine.tsx
```diff
@@ -0,0 +1,52 @@
...
+          target: "notFound",
+        },
+      },
+    },
+    notFound: {
+      on: {
+        "reconnect.retry": "retrying",
+      },
+      entry: ["notifyNotFound"],
+    },
```
phryneas, line 50: Previously, this also had a `timedout` state, but we never triggered it from anywhere.

src/application/machines/reconnectMachine.tsx
```diff
@@ -0,0 +1,52 @@
...
+    retrying: {
+      entry: ["notifyWaitingForConnection"],
+      after: {
+        connectTimeout: {
+          target: "notFound",
+        },
+      },
+    },
+    notFound: {
```
phryneas, line 45: The modal is now shown while the machine is in this state. If we wanted to show the modal in more states that just one, we could look into `tag`.

src/application/machines/reconnectMachine.vscode.tsx
```diff
@@ -0,0 +1,18 @@
+import { setup } from "xstate";
+export const reconnectMachine = setup({
+  delays: { connectTimeout: 500 },
+}).createMachine({
```
phryneas, line 4: In VSCode, we want to show this from the start, as it has valuable information.
We will wait for a very short amount of time though, in case something immediately connects - that way we avoid flickering of the modal.

src/application/components/ExternalLink.tsx
```diff
@@ -0,0 +1,8 @@
+import { type ComponentProps, forwardRef } from "react";
+
+export const ExternalLink = forwardRef<
+  HTMLAnchorElement,
+  ComponentProps<"a"> & { href: string }
+>((props, ref) => {
+  return <a {...props} ref={ref} />;
```
jerelmiller, line 7: Since we are moving to this, let's add `target="_blank" rel="noopener noreferrer"` as props here as well since external links on the web side don't open unless they are in a new tab.

src/application/components/GitHubReleaseHoverCard.tsx
```diff
@@ -68,27 +69,35 @@ function SnapshotCardContents({ version }: { version: SnapshotVersion }) {
             <div className="flex gap-1 items-center text-xs font-bold uppercase text-secondary dark:text-secondary-dark">
               Published {formatDate(parseSnapshotTimestamp(release.timestamp))}
             </div>
-            <a
+            <ExternalLink
               className="flex gap-1 items-center mt-2"
               href={`https://github.com/apollographql/apollo-client/pull/${release.prNumber}`}
               target="_blank"
               rel="noreferrer"
```
jerelmiller, line 76:
```suggestion
```

Can we remove these props once the browser `ExternalLink` is updated to add these props directly?

phryneas:
Okay, I went over that - re-review please :)

src/application/machines/reconnectMachine.tsx
jerelmiller: I'm glad you got rid of it. It was a relic from the days before we had the client send a register event. Devtools used to ping the content scripts, so the `timedout` event was used if devtools couldn't communicate with the content scripts anymore, but that is largely a thing of the past.

src/application/App.tsx
```diff
@@ -150,26 +135,22 @@ export const App = () => {
...
 
   return (
     <>
       <SettingsModal open={settingsOpen} onOpen={setSettingsOpen} />
-      <ClientNotFoundModal
-        open={snapshot.context.modalOpen}
-        onClose={() => send({ type: "closeModal" })}
-        onRetry={() => send({ type: "retry" })}
-      />
+      <Modals />
```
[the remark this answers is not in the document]
jerelmiller: Can we name this something like `ErrorModals`? Ideally we have some namespace that describes that these are the modals opened when the clients can't connect in some way. 

Otherwise it looks a little bit weird to have `SettingsModal` right next to `Modals` 😂.

src/application/machines/devtoolsMachine.tsx
```diff
@@ -1,143 +1,233 @@
...
   },
   on: {
-    closeModal: {
-      actions: "closeModal",
+    // forward reconnect events to child actor
+    "reconnect.*": {
+      actions: sendTo("reconnect", ({ event }) => event),
+    },
+    // on an `emit.*` event, `emit` that event so it can be subscribed to from the app
+    "emit.*": {
```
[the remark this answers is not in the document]
phryneas: Before https://github.com/apollographql/apollo-client-devtools/pull/1536/files, this would allow us to call `refetch` as a reaction to that event.

It's not like we need this anymore *right now*, but it is a good blueprint we can build on if we need something like this in the future, and it doesn't hurt, so I'd keep it in for now.

src/application/components/Modals/PortNotOpenModal.vscode.tsx
```diff
@@ -0,0 +1,148 @@
...
+>();
+
+interface ErrorModalProps {
+  open: boolean;
+}
+
+const StartCommandLabel = "Apollo: Start Apollo Client DevTools Server";
+const StartCommand = "apollographql/startDevToolsServer";
+const StopCommandLabel = "Apollo: Stop Apollo Client DevTools Server";
+const PortSetting = "apollographql.devTools.serverPort";
```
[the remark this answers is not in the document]
phryneas: 4679bd92bac60a3a92590da1765adbec95dad6cc ✅

jerelmiller (approved):
Went over this in person.

Only feedback I have left is to rename that `Modals` folder (https://github.com/apollographql/apollo-client-devtools/pull/1513#discussion_r1873994093), but I'll trust you'll do that.

Looking forward to getting this out!

PR #1520, opened 2024-09-16
wrap `injectScript` call in `try..catch`, offer extension builds as PR artifacts

### LINKED ISSUE #1519: Extension in a default-src 'self' CSP environment causes page load to fail
**Intended outcome:**

Load a page with a CSP of `default-src 'self';` (in this case https://standardebooks.org/). The page is expected to load and render regardless of the CSP settings.

**Actual outcome:**

Blank screen, and CSP errors in the console. This was originally reported at https://github.com/standardebooks/web/issues/397.

Visit https://standardebooks.org/, observe blank page, observe errors in the console.

- OS: [e.g. iOS] macOS 14.6.1
- Browser [e.g. chrome, safari] Firefox
- Browser version [e.g. 22] I tested in Nightly
- Extension version [e.g. 3.0] 4.18.6

phryneas:
This *might* resolve #1519.

Also adding a "download build extension" comment to future PRs with this, which would make it easier to get quick user feedback.

### DIFF
--- a/src/extension/tab/tab.ts
+++ b/src/extension/tab/tab.ts
@@ -50,5 +50,12 @@ if (__IS_FIREFOX__) {
     }
   }
 
-  injectScript(browser.runtime.getURL("hook.js"));
+  try {
+    injectScript(browser.runtime.getURL("hook.js"));
+  } catch (e) {
+    console.warn(
+      "Error occured while trying to initialize Apollo Client Devtools:",
+      e
+    );
+  }
 }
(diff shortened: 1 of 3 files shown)


src/extension/tab/tab.ts
```diff
@@ -50,5 +50,12 @@ if (__IS_FIREFOX__) {
...
 
-  injectScript(browser.runtime.getURL("hook.js"));
+  try {
+    injectScript(browser.runtime.getURL("hook.js"));
+  } catch (e) {
+    console.warn(
+      "Error occured while trying to initialize Apollo Client Devtools:",
+      e
+    );
+  }
```
phryneas, line 60: This doesn't work, as the error is triggered delayed when the script evaluates, not when it is inserted in DOM.

Also, listening for a `securitypolicyviolation` event allows me to read the details of the violation, but not to prevent it from crashing the whole page. I'm a bit out of ideas.

PR #1607, opened 2024-12-20
Add memory panel to monitor internal caches

jerelmiller:
Adds a new "Memory" panel to the devtools that shows the current values of the internal caches.

Screenshot 2024-12-20 at 3 33 44 PM
Screenshot 2024-12-20 at 3 15 43 PM
Screenshot 2024-12-20 at 3 15 49 PM


jerelmiller:
Possible naming alternatives if we decide not to call this memory.
- Memoization limits
- Resource usage

development/client/vite.config.ts
```diff
@@ -2,6 +2,9 @@ import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
 
 export default defineConfig({
+  define: {
+    "globalThis.__DEV__": JSON.stringify(true),
+  },
```
phryneas, line 7: I think this shouldn't be necessary anymore.

development/client/src/ColorSchemeGenerator.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect } from "react";
-import { useQuery } from "@apollo/client/react";
+import { useQuery } from "./ClientContext";
```
phryneas, line 2: Good catch!

phryneas:
Apollo Client in the dev app needs an update to a recent version.

src/application/components/CacheSize.tsx
```diff
@@ -0,0 +1,305 @@
...
+      <p>
+        Cache size for the <code>executeSelectionSet</code> method on{" "}
+        <TooltipLink href="https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/readFromStore.ts">
+          <code>StoreReader</code>
+        </TooltipLink>
+        .
+      </p>
+      <p>
+        Note: <code>executeSelectionSet</code> will be set to the{" "}
+        <code>resultCacheMaxSize</code> option and will fall back to this
```
phryneas, line 192: We removed `resultCacheMaxSize` in 4.0, so this note should be removed or displayed conditionally (here and in 2 more places).

src/application/components/MemoryInternals.tsx
```diff
@@ -0,0 +1,295 @@
...
+            >
+              docs
+              <IconOutlink className="size-4" />
+            </a>
+            .
+          </p>
+        </header>
+
+        <ButtonGroup>
+          <Tooltip content="View historical">
```
phryneas, line 207: I'm not sure - is "historical" the right description here, or should it be something like "as dashboard"?

phryneas (approved):
This looks great to me, thank you for getting this over the finish line :)

jerelmiller:
Ya I think I like memoization better. I fear "memory" might make people think we are tracking bytes consumed, which obviously isn't the feature here so let's do that!

src/application/components/CacheSize.tsx
jerelmiller: Seems we have an out-of-date doc-block in Apollo Client itself since I got this straight from there. I'll open a PR over there to fix. Eventually it would be nice to have this come straight from the doc block itself.

PR #1694, opened 2025-08-26
shared AC3/AC4 dev client app

phryneas:
As a suggestion for the 4.0 PR - WDYT about having one share AC3/AC4 app?

This would cause problems with more complicated hook usage, but since they are similar enough, running the AC3 hooks with the AC4 types should be fine.


jerelmiller (approved):
This is probably the right call. Thanks!

PR #1713, opened 2025-09-05
fix: check if handler exist before sending tab command

braineo:
The devtool monkey patches the stop method, however the stop method can be called in react strict mode twice. On the second call, the Map does not have a handler and leads to reading id of undefined error.

When it happens, it crashes the react app and creates a lot confusions for bug tracking.

This change enables back the type safety and make sure id exists before calling clientTerminated

### DIFF
--- a/.changeset/spicy-trees-tease.md
+++ b/.changeset/spicy-trees-tease.md
@@ -1,0 +1,6 @@
+---
+"apollo-client-devtools": patch
+"@apollo/client-devtools-vscode": patch
+---
+
+fix calling client.stop twice causing app crashing by checking if handler exist before sending tab command


braineo:
Hi @jerelmiller , the bug was introduced in commit https://github.com/apollographql/apollo-client-devtools/commit/bd953e0e66ad37eb57a7c3e2297ea5583534a538

would you be able to take a quick look at this pr? my colleagues had to disable the devtool for development and it was a surprise that the uncaught error was from the devtool. Thanks!

jerelmiller:
Good catch @braineo! Would you mind adding a changeset (`npx changeset`)? I'd be happy to merge this in and get this released after that 🙂

PR #1722, opened 2025-09-18
Fix explorer infinite loading when errors are returned

### LINKED ISSUE #1721: Apollo 4: Errors don't show up in explorer and instead infinite loads
**Intended outcome:**

If the backend returns an error, the explorer should show the error the same as with Apollo 3.
Expected result:
```json
{
  "error": {},
  "errors": [
    {
      "message": "Internal error",
      "path": [
        "search"
      ]
    }
  ]
}
```

**Actual outcome:**

Instead, the explorer infinitely loads and never shows anything in the result pane whenever an error is thrown on the backend.

Throw an error in the backend and try to execute a query through the devtools explorer.
Make sure you're on Apollo 4 client.
This is the response of the backend that Apollo 4 devtools doesn't want to load, which work fine with Apollo 3:
```
{"errors":[{"message":"Internal error","path":["search"]}],"data":null}
```

- OS: MacOS Sequoia 15.6
- Browser: Zen (Firefox fork)
- Browser version: 1.15.5b (Firefox 142.0.1) (aarch64)
- Extension version:  4.21.4

jerelmiller:
Fixes #1721

We were filtering the response from the client based on `dataState` being non-empty, but this meant we were filtering out responses that returned errors since `errorPolicy: none` sets `data` to `undefined`. This fix will ensure the explorer response message is returned when the network status has settled.

### DIFF
--- a/.changeset/yellow-eyes-look.md
+++ b/.changeset/yellow-eyes-look.md
@@ -1,0 +1,5 @@
+---
+"apollo-client-devtools": patch
+---
+
+Fix an issue where executing a query in the explorer tab that returned errors would get stuck in the loading state indefinitely.
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
       ],
       "dependencies": {
         "@apollo/brand": "^0.7.0",
-        "@apollo/client": "4.0.3",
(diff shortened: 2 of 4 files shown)


phryneas:
Post-approved ✅

PR #1746, opened 2025-11-21
Show a banner instead of crashing the page when the devtools can't communicate with the content scripts

jerelmiller:
Screenshot 2025-11-21 at 3 02 50 PM


.changeset/clever-planets-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"apollo-client-devtools": minor
+---
+
+Show a banner with a prompt to restart the page and the devtools instead of crashing the page when the extension can't communicate with the content scripts .
```
jerelmiller:
```suggestion
Show a banner with a prompt to restart the page and the devtools instead of crashing the page when the extension can't communicate with the content scripts.
```

PR #1800, opened 2026-01-29
run introspection queries lazy and handle resulting errors better

phryneas:
This fixes two things:

* Previously, `executeQuery` would call `client.query(` without specifying an `errorPolicy`, which could cause the promise to reject - our logic expected it to resolve with the `error`, though. This did lead to a number of uncaught promise rejections.
* Previously, the introspection query was fired off immediately when the extension started, not when the user moved into the explorer tab. As a result, users could have errors ending up in their link chain the moment you started the devtools if the server had introspection disabled. We can't avoid that error in the linkchain fully, but we can at least defer when the introspection query runs. Since it's not needed before opening the Explorer window, it's now deferred until the explorer is visible.

Also, the whole Explorer iframe logic was very complicated and added a lot of code to deal with the `null`ishness of the iframe - which only played a role in the initial render and would become irrelevant already on the second render. I moved this to an `useImperativeHandle`-managed ref instead.

Some of these changes also required a TS version bump, so I did that too.

I also bumped the ESLint rules for react-hooks by 3 major versions (the current ones didn't know `useEffectEvent`) and also removed that other non-official `eslint-plugin-react` since it seems seriously outdated.


src/application/components/Explorer/Explorer.tsx
```diff
@@ -174,24 +187,40 @@ export const Explorer = ({
...
+  });
+
+  const runIntrospectionQuery = useEffectEvent(() => {
+    if (!clientId) {
+      return;
+    }
+    const last = lastIntrospection.current;
+    if (last && last.clientId === clientId) {
+      return;
+    }
```
jerelmiller: Since you're already checking the truthiness of `clientId` in the early return above, you should be able to simplify and remove the check for `if (last)` since you'll never compare a non-truthy value to a non-truthy value here (at least I assume you were trying to guard against `undefined === undefined` with `if (last)`).

```suggestion
    if (lastIntrospection.current?.clientId === clientId) {
      return;
    }
```

jerelmiller (approved):

PR #1859, opened 2026-04-24
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `src/application/components/HighlightMatch.tsx:L7`

`HighlightMatch` builds a `RegExp` directly from `searchTerm` (`new RegExp(searchTerm, "i")`). If `searchTerm` is user-controlled, attackers can supply regex metacharacters to trigger expensive backtracking (ReDoS) or runtime exceptions (invalid regex), causing UI freezes or crashes.

## Solution

Escape user input before creating a regex (e.g., `searchTerm.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")`), enforce max input length, and wrap regex creation in `try/catch` to prevent crashes.

## Changes

- `src/application/components/HighlightMatch.tsx` (modified)
- `src/application/components/ObjectViewer/ObjectKey.tsx` (modified)

### DIFF
--- a/src/application/components/HighlightMatch.tsx
+++ b/src/application/components/HighlightMatch.tsx
@@ -4,7 +4,20 @@ interface HighlightMatchProps {
 }
 
 const HighlightMatch = ({ searchTerm, value }: HighlightMatchProps) => {
-  const regex = new RegExp(searchTerm, "i");
+  if (!searchTerm) {
+    return <>{value}</>;
+  }
+
+  const escapedSearchTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+  let regex: RegExp;
+
+  try {
+    regex = new RegExp(escapedSearchTerm, "i");
+  } catch {
+    return <>{value}</>;
+  }
+
   const match = regex.exec(value);
 
   if (!match) {
@@ -17,7 +30,7 @@ const HighlightMatch = ({ searchTerm, value }: HighlightMatchProps) => {
       <span className="bg-searchHighlight dark:bg-searchHighlight-dark text-inverted dark:text-inverted-dark">
         {match[0]}
       </span>
-      {value.slice(match.index + searchTerm.length)}
+      {value.slice(match.index + match[0].length)}
     </span>
   );
 };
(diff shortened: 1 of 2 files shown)


src/application/components/ObjectViewer/ObjectKey.tsx
```diff
@@ -14,7 +14,13 @@ export const ObjectKey = customRenderable(
   ({ className, value, softWrapCharacters }: ObjectKeyProps) => {
     const uniqueChars = new Set(softWrapCharacters);
     const regex = softWrapCharacters
-      ? new RegExp(`(${softWrapCharacters.join("|")})`)
+      ? new RegExp(
+          `(${softWrapCharacters
+            .map((character) =>
+              character.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")
```
jerelmiller: I'd love to use [`RegExp.escape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape) if we could. That supports Chrome and Firefox versions that have been out for a year so I think its pretty safe to use it at this point in the extension. Can you update these to `RegExp.escape`?


```suggestion
              RegExp.escape(character)
```

src/application/components/HighlightMatch.tsx
```diff
@@ -4,7 +4,20 @@ interface HighlightMatchProps {
 }
 
 const HighlightMatch = ({ searchTerm, value }: HighlightMatchProps) => {
-  const regex = new RegExp(searchTerm, "i");
+  if (!searchTerm) {
+    return <>{value}</>;
+  }
+
+  const escapedSearchTerm = searchTerm.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
jerelmiller:
```suggestion
  const escapedSearchTerm = RegExp.escape(searchTerm);
```

jerelmiller (changes requested):
I'd love to use `RegExp.escape`, otherwise this looks good.

Before I accept this, can you sign the CLA please?

